PR #7, opened 2016-02-24
stop acquisition on schema violations

frasertweedale:
Currently there is no way to distinguish between "incorrect
passphrase" and obvious metadata schema violations such as missing
keys, invalid/unsupported algorithms, or invalid base64-encoded
data.  Acquisition loops forever in these scenarios.

Add the `decrypt_result` enum and `clevis_decrypt_result_t` type to
distinguish success, unrecoverable errors and potentially
recoverable errors in the result of `crypto_decrypt`.

Note that corrupted "iter", "ct" and "salt" values that are still
valid integer or base64 data will not be detected even after this
change.

### DIFF
--- a/src/clevis.h
+++ b/src/clevis.h
@@ -30,17 +30,29 @@ typedef struct clevis_acquire_f clevis_acquire_f;
 typedef struct clevis_buf_f clevis_buf_f;
 typedef struct clevis_pin_f clevis_pin_f;
 
+typedef struct clevis_decrypt_result_t clevis_decrypt_result_t;
 typedef struct clevis_pwd_t clevis_pwd_t;
 typedef struct clevis_buf_t clevis_buf_t;
 
 typedef bool clevis_pwd_vfy(const clevis_buf_t *pwd, clevis_pwd_t *misc);
+
+enum decrypt_result {
+  DECRYPT_SUCCESS,
+  DECRYPT_FAIL_STOP,
+  DECRYPT_FAIL_TRYAGAIN
+};
+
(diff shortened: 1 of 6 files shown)


frasertweedale:
Note that this implementation will proceed with acquiring a key (e.g. via password prompt or tang)
*before* calling `decrypt` and realising there is a problem.  There is no way to _force_ a pin to
check `cipher`/`kdf`/`salt`/`iter`/`ct` schema validity before doing it's thing.  However, the following pattern might make it easy for pin authors avoid doing extra work when decryption is certain to fail:

1. provide a new struct containing decoded encryption params:

        typedef struct clevis_decrypt_params {
            EVP_CIPHER *cipher;
            EVP_MD *kdf;
            clevis_buf_t salt;   // decoded salt
            int iter;
            clevis_buf_t ct;     // decoded ciphertext
        } clevis_decrypt_params_t;

2. provide another function in `clevis_acquire_f` for "reading" encryption params out of the given JSON:

        clevis_decrypt_params_t *(*decrypt_params)(const json_t *data);

3. update signature of `decrypt` to take a `clevis_decrypt_params_t` instead of a full `json_t`.

        clevis_decrypt_result_t (*decrypt)(const clevis_buf_t *key, clevis_decrypt_params_t *params);

4. pin implementation is then: 1) call `decrypt_params` and fail early if it returns a bad value; 2) acquire outer key; 3) decrypt using outer key and params retrieved at (1).

frasertweedale:
Closing stale PR.

PR #18, opened 2017-08-11
Add Initramfs support

ShaRose:
I was going to wait until I updated the build script, but I clearly haven't gotten around to it. It's fairly well tested and shouldn't have any 'gotchas': I've been using it at home for the past 2 months without issue on a few systems.

/src/initramfs/* should be copied to /usr/share/initramfs-tools/, assuming initramfs is used instead of dracut.


jh23453:
I've used the script as a start to get clevis running on debian. Here are the changes I did:

```
--- clevis      2018-02-10 18:26:28.517508839 +0100
+++ /etc/initramfs-tools/scripts/local-top/clevis       2018-02-10 20:24:31.191136875 +0100
@@ -39,9 +39,8 @@
 # Print the PID of the askpass process with a file descriptor opened to
 # /lib/cryptsetup/passfifo if there is one.
 get_askpass_pid() {
-    psinfo=$(ps) # Doing this so I don't end up matching myself
-    echo "$psinfo" | awk "/$cryptkeyscript/ { print \$1 }" | \
-        while read pid; do
+    local pid
+    for pid in $(pidof "askpass" "plymouth"); do
         if in_fds "$pid" "$PASSFIFO"; then
             echo "$pid"
             break
@@ -59,13 +58,6 @@
     # Set the path how we want it (Probably not all needed)
     PATH="/sbin:/usr/sbin:/bin:/usr/bin:/usr/local/bin"
 
-    if [ -x /bin/plymouth ] && plymouth --ping; then
-        cryptkeyscript='plymouth ask-for-password'
-    else
-        # This has to be escaped for awk
-        cryptkeyscript='\/lib\/cryptsetup\/askpass'
-    fi
-
     PASSFIFO='/lib/cryptsetup/passfifo'
 
     OLD_CRYPTTAB_SOURCE=""
@@ -80,7 +72,7 @@
         done
 
         # Import CRYPTTAB_SOURCE from the askpass process.
-        local $(grep '^CRYPTTAB_SOURCE=' /proc/$pid/environ)
+        local $(tr '\0' '\n' </proc/$pid/environ | grep '^CRYPTTAB_SOURCE=' )
 
         # Make sure that CRYPTTAB_SOURCE is actually a block device
         [ ! -b "$CRYPTTAB_SOURCE" ] && continue
```
The first two hunks use pidof to get the process id of askpass. I think it's more readable
and reliable.

The last hunk fixes parsing /proc/$pid/environment. In that file the variable=value pairs 
are delimited with \\0, so "^CRYPTTAB_SOURCE=" won't match most of the time - only
if it's the first entry in the environment. So let's fix it: translate \\0 to \\n and now we can match
"^CRYPTTAB_SOURCE=".

Thanks for the scripts - they are really useful for me.

PR #24, opened 2017-11-22
Add clevis-luks-{list,unbind} commands

martinezjavier:
This pull request adds two new clevis commands for managing bound LUKS volumes. The clevis-luks-list command lists the pins bound to a LUKS volume and the clevis-luks-unbind command unbinds a pin bound to a LUKS volume.

While being there, fix a trivial typo in the man pages.


martinezjavier:
@npmccallum any comments on this?

npmccallum:
I have merged the first patch. The other two need to be split into separate pull requests.

The unbind command is close to merge. Do we want to have a confirmation (and corresponding -f option) before performing the destructive action? Also, I'm not sure we want to fail the whole process if the cryptsetup call fails. Perhaps we should check that the slot is active before we try to kill it?

The list command is a bit less than I want. I'm not sure how helpful listing the top level pin is in a recursive configuration structure. I think instead we want each pin to define a command to reconstitute its configuration from the JWE. Then, we can print the configuration for the slot by recursively calling this command.

martinezjavier:
@npmccallum thanks a lot for the review, I'll try to find time to re-spin this PR addressing the issues you pointed out.

PR #25, opened 2018-02-19
[v2] Add clevis-luks-unbind command

martinezjavier:
This is a v2 of PR #24 that addresses the issues pointed out by @npmccallum.

Changes since v1:

- Ask for confirmation before wiping the luksmeta and LUKSv1 slots and provide a `-f` option to force it.
- First wipe the luksmeta slot and then the LUKSv1 one.
- Don't attempt to wipe the LUKSv1 slot if isn't active.


src/clevis-luks-unbind
```diff
@@ -0,0 +1,96 @@
...
+    echo "$SUMMARY"
+    exit 0
+fi
+
+while getopts ":d:s:f" o; do
+    case "$o" in
+    f) FRC=-q;;
+    d) DEV=$OPTARG;;
+    s) SLT=$OPTARG;;
+    *) usage;;
```
npmccallum: We probably shouldn't continue running the program in this case.

src/clevis-luks-unbind
```diff
@@ -0,0 +1,96 @@
...
+    echo "Did not specify a slot!" >&2
+    usage
+fi
+
+if ! luksmeta test -d $DEV 2>/dev/null; then
+    echo "The $DEV device is not valid!" >&2
+    exit 1
+fi
+
+read -r slot active uuid <<< $(luksmeta show -d "$DEV" | grep "$SLT")
```
npmccallum: The `grep` here is wrong. It will match numbers 1-8 in the UUID.

src/clevis-luks-unbind
```diff
@@ -0,0 +1,96 @@
...
+
+if ! luksmeta wipe -f -d "$DEV" -u "$UUID" -s "$SLT"; then
+    echo "LUKSMeta slot $SLT for device $DEV couldn't be deleted"
+    exit 1
+fi
+
+if [ "$active" != "active" ]; then
+   echo "LUKSv1 slot $SLT on device $DEV is not active." >&2
+   exit 1
+fi
```
npmccallum: I think this check needs to be merged with the next one. Namely:
```
if [ "$active" == "active" ] && ! cryptsetup luksKillSlot "$DEV" "$SLT" "$FRC"; then
```

src/clevis-luks-unbind
martinezjavier: We won't since the usage() function exists with a status code 1.

src/clevis-luks-unbind
martinezjavier: I don't think so, because the slot not being active isn't an error, it just means that we have a LUKSMeta slot with no corresponding LUKSv1 slot. But failing to kill a LUKSv1 slot is an error and probably we shouldn't continue in this case.

martinezjavier:
@npmccallum I pushed a new version that hopefully addresses the issues you pointed out. Please let me know what you think.

src/clevis-luks-unbind
```diff
@@ -0,0 +1,94 @@
...
+
+read -r slot active uuid <<< $(luksmeta show -d "$DEV" | grep "^$SLT *")
+
+if [ "$uuid" = "empty" ]; then
+   echo "The LUKSMeta slot $SLT on device $DEV is already empty." >&2
+   exit 1
+fi
+
+if [ "$active" = "active" ]; then
+    if ! cryptsetup luksKillSlot "$DEV" "$SLT" "$FRC"; then
```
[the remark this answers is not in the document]
martinezjavier: @npmccallum nice catch, you are right. I fixed that too.

PR #33, opened 2018-02-27
Make build easier

martinezjavier:
This pull request adds a bootstrap script so users don't need to manually call autoreconf with the correct options and an INSTALL.md file that explains how to build from source and the project dependencies.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,80 @@
+This file contains instructions to build and install Clevis from source
+
+# Dependencies
+To build and install the Clevis software the following software packages
+are required. In many cases dependencies are platform specific and so the
+following sections describe them for the supported platforms.
+
+## Linux:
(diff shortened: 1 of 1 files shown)


martinezjavier:
I just noticed that there's a sectio nin the README.md that asks Clevis to not be built from source and instead use the distribution packages. But I still think that it would be good to provide this info for people that don't have other option (i.e: clevis not packaged in their distro, not using distros but instead building images using BitBake / OE, etc).

My motivation for these patches is that several people asked me questions about Clevis build and install, due my post on the tpm2 pin.

npmccallum:
@martinezjavier I agree. In fact, I originally had this working with meson but ported to autotools. We should keep it as a goal for the future.

Phant0mas:
Thank you for adding a list of dependencies.

PR #37, opened 2018-03-06
Don't execute clevis-luks-unlock while loop in a subshell

martinezjavier:
The loop that tries to open the dm-crypt devices using the pins in the
luksmeta header is executed in a subshell. So on success it calls exit
to exit the subshell.

But then clevis-luks-unlock has no way to know if the encrypted device
was opened correctly or not. So run the loop in the main shell process
and return 0 as exit status if the operation was successful.

Fixes: #36

### DIFF
--- a/src/clevis-luks-unlock
+++ b/src/clevis-luks-unlock
@@ -54,7 +54,7 @@ fi
 
 NAME=${NAME:-luks-`cryptsetup luksUUID $DEV`}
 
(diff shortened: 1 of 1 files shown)


martinezjavier:
@npmccallum Ok, I did the change that you requested.

npmccallum:
Thanks!

PR #41, opened 2018-03-23
Add bash-completion support

martinezjavier:
This patch adds bash completion support for clevis, to allow
the shell autocomplete the clevis commands and their options.


npmccallum:
If I understand it correctly, this patch hard codes the completions rather than locating them in the filesystem. This has two problems as I see it. First, it means that we have to update this every time we add a new command; which is error-prone. Second, it means that if someone else installs a command we can't auto-complete it.

Is it possible to furnish the autocompletions from the filesystem itself?

martinezjavier:
@npmccallum good point, I used as inspiration other bash completion scripts that were hard coding their options but I agree that isn't flexible and error prone. I'll try to find some time to rework the script.

Thanks a lot for your feedback.

martinezjavier:
Yes, I'll look at this next week.

PR #47, opened 2018-04-24
Fix decryption of http with jwk+json type

mvollmer:
The expected "jwk" variable wasn't set in this case and the actual
decryption would fail with "Invalid key!".


mvollmer:
(Upps, sorry about the typo in the branch name. I guess it doesn't matter.)

npmccallum:
Thanks!

PR #49, opened 2018-05-03
Check for compatible tpm2-tools version

martinezjavier:
The tpm2 pin requires tpm2-tools version 3, so check if the correct
version was installed and exit with an error if that's not the case.

Fixes: #48


martinezjavier:
@npmccallum any comments about this PR? The change is really trivial.

PR #56, opened 2018-05-28
tpm2: Replace -H option, use --hierarchy!

gitfineon:
Option -H is outdated and replaced by -a/--hierarchy and there is a
default without this option. Use default to keep it simple!
https://github.com/tpm2-software/tpm2-tools/commit/504289ddb6495b25c2fa5ed8aa125617e9bba223

### DIFF
--- a/src/clevis-decrypt-tpm2
+++ b/src/clevis-decrypt-tpm2
@@ -18,9 +18,6 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-# The owner hierarchy is the one that should be used by the Operating System.
-auth="o"
-
 function on_exit() {
     if ! rm -r $TMP; then
         echo "Delete temporary files failed!" >&2
@@ -37,8 +34,16 @@ if [ -t 0 ]; then
     exit 1
 fi
 
-export TPM2TOOLS_TCTI_NAME=device
-export TPM2TOOLS_DEVICE_FILE=`ls /dev/tpmrm? 2>/dev/null`
+TPM2TOOLS_INFO=`tpm2_pcrlist -v`
+TPM2TOOLS_DEVICE_FILE=`ls /dev/tpmrm? 2>/dev/null`
+if [[ $TPM2TOOLS_INFO = *version=\"3.* ]]; then
+    export TPM2TOOLS_TCTI_NAME=device
+    export TPM2TOOLS_DEVICE_FILE
+    export TPM2TOOLS_TCTI="device:$TPM2TOOLS_DEVICE_FILE"
+else
+    echo "The tpm2 pin requires tpm2-tools version 3 or higher" >&2
+    exit 1
+fi
 
 if [ -z "${TPM2TOOLS_DEVICE_FILE[0]}" ]; then
     echo "A TPM2 device with the in-kernel resource manager is needed!" >&2
@@ -89,6 +94,23 @@ fi
 
 trap 'on_exit' EXIT
 
+if [[ $TPM2TOOLS_INFO = *version=\"3.* ]]; then
+    # The (o)wner hierarchy is the one that should be used by the OS.
+    # For tpm-tools version 3.X explicit hierarchy parameter is required,
+    # but should be optional in the future.
+    #arg_createprimary="-H o -C $TMP/primary.context"					# short 3.X
+    arg_createprimary="--hierarchy o --context $TMP/primary.context"			# long  3.X
+    #arg_load="-c $TMP/primary.context -C $TMP/load.context"				# short 3.X
+    arg_load="--context-parent $TMP/primary.context --context $TMP/load.context"	# long  3.X
+else
+    echo "Developer flag: Use at your own risk!"
+    arg_createprimary="--hierarchy o --out-context $TMP/primary.context"		# long  4.X
+    arg_load="--context-parent $TMP/primary.context --out-context $TMP/load.context"	# long  4.X
+fi
+
+#arg_unseal="-c $TMP/load.context"		# short 3.X
+arg_unseal="--item-context $TMP/load.context"	# long  3.X
+
 pcr_ids=`jose fmt -j- -Og clevis -g tpm2 -g pcr_ids -Su- <<< "$jhd"` || true
 
 if [ -n "$pcr_ids" ]; then
@@ -106,19 +128,28 @@ if ! `jose b64 dec -i- -O $TMP/jwk.priv <<< "$jwk_priv"`; then
     exit 1
 fi
 
-if ! tpm2_createprimary -Q -H "$auth" -g "$hash" -G "$key" \
-     -C $TMP/primary.context 2>/dev/null; then
+ls -la $TMP
+echo "tpm2_createprimary -Q -g "$hash" -G "$key" $arg_createprimary; then"
+
+if ! tpm2_createprimary -Q -g "$hash" -G "$key" \
+     $arg_createprimary; then
     echo "Creating TPM2 primary key failed!" >&2
     exit 1
 fi
 
-if ! tpm2_load -Q -c $TMP/primary.context -u $TMP/jwk.pub -r $TMP/jwk.priv \
-     -C $TMP/load.context 2>/dev/null; then
+ls -la $TMP
+echo "tpm2_load -Q -u $TMP/jwk.pub -r $TMP/jwk.priv $arg_load; then"
+
+if ! tpm2_load -Q -u $TMP/jwk.pub -r $TMP/jwk.priv \
+     $arg_load; then
     echo "Loading jwk to TPM2 failed!" >&2
     exit 1
 fi
 
-if ! jwk=`tpm2_unseal -c $TMP/load.context $policy_options 2>/dev/null`; then
+ls -la $TMP
+echo "tpm2_unseal -c $TMP/load.context $policy_options then"
+
+if ! jwk=`tpm2_unseal $arg_unseal $policy_options`; then
     echo "Unsealing jwk from TPM failed!" >&2
     exit 1
 fi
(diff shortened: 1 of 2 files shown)


gitfineon:
Hey @martinezjavier, please ack!

martinezjavier:
@gitfineon yes, I've a similar patch but I haven't proposed yet since there isn't a released version of the tpm2-tools with that change. But we also need to take into account for tpm2-tools 3.x and any newer version that removed this option. Also, your patch doesn't take into account that the env vars for the TCTI have changed.

The patch I've is the following:

```
commit 8472474227a2acc7e73000e8e6c0920670c8226c
Author: Javier Martinez Canillas <javierm@redhat.com>
Date:   Wed May 2 11:24:22 2018 +0200

    Allow tpm2 pin to be used with both tpm2-tools version 3 and 4
    
    The tpm2-tools version 4 changed some of the tools options and environment
    variables used, so check the tpm2-tools version and take this into account.
    
diff --git a/src/clevis-decrypt-tpm2 b/src/clevis-decrypt-tpm2
index f3871d83d81..5c509d69ca0 100755
--- a/src/clevis-decrypt-tpm2
+++ b/src/clevis-decrypt-tpm2
@@ -37,8 +37,17 @@ if [ -t 0 ]; then
     exit 1
 fi
 
-export TPM2TOOLS_TCTI_NAME=device
-export TPM2TOOLS_DEVICE_FILE=`ls /dev/tpmrm? 2>/dev/null`
+TPM2TOOLS_INFO=`tpm2_pcrlist -v`
+TPM2TOOLS_DEVICE_FILE=`ls /dev/tpmrm? 2>/dev/null`
+
+if [[ $TPM2TOOLS_INFO = *version=\"3.* ]]; then
+    hierarchy_opt="-H"
+    export TPM2TOOLS_TCTI_NAME=device
+    export TPM2TOOLS_DEVICE_FILE
+else
+    hierarchy_opt="-a"
+    export TPM2TOOLS_TCTI="device:$TPM2TOOLS_DEVICE_FILE"
+fi
 
 if [ -z "${TPM2TOOLS_DEVICE_FILE[0]}" ]; then
     echo "A TPM2 device with the in-kernel resource manager is needed!" >&2
@@ -106,7 +115,7 @@ if ! `jose b64 dec -i- -O $TMP/jwk.priv <<< "$jwk_priv"`; then
     exit 1
 fi
 
-if ! tpm2_createprimary -Q -H "$auth" -g "$hash" -G "$key" \
+if ! tpm2_createprimary -Q "$hierarchy_opt" "$auth" -g "$hash" -G "$key" \
      -C $TMP/primary.context 2>/dev/null; then
     echo "Creating TPM2 primary key failed!" >&2
     exit 1
diff --git a/src/clevis-encrypt-tpm2 b/src/clevis-encrypt-tpm2
index b99aa97f7f6..f7a5b2f2368 100755
--- a/src/clevis-encrypt-tpm2
+++ b/src/clevis-encrypt-tpm2
@@ -59,8 +59,17 @@ if [ -t 0 ]; then
     exit 1
 fi
 
-export TPM2TOOLS_TCTI_NAME=device
-export TPM2TOOLS_DEVICE_FILE=`ls /dev/tpmrm? 2>/dev/null`
+TPM2TOOLS_INFO=`tpm2_pcrlist -v`
+TPM2TOOLS_DEVICE_FILE=`ls /dev/tpmrm? 2>/dev/null`
+
+if [[ $TPM2TOOLS_INFO = *version=\"3.* ]]; then
+    hierarchy_opt="-H"
+    export TPM2TOOLS_TCTI_NAME=device
+    export TPM2TOOLS_DEVICE_FILE
+else
+    hierarchy_opt="-a"
+    export TPM2TOOLS_TCTI="device:$TPM2TOOLS_DEVICE_FILE"
+fi
 
 if [ -z "${TPM2TOOLS_DEVICE_FILE[0]}" ]; then
     echo "A TPM2 device with the in-kernel resource manager is needed!" >&2
@@ -99,7 +108,7 @@ fi
 
 trap 'on_exit' EXIT
 
-if ! tpm2_createprimary -Q -H "$auth" -g "$hash" -G "$key" -C $TMP/primary.context; then
+if ! tpm2_createprimary -Q "$hierarchy_opt" "$auth" -g "$hash" -G "$key" -C $TMP/primary.context; then
     echo "Creating TPM2 primary key failed!" >&2
     exit 1
 fi
```

martinezjavier:
@gitfineon the version check is something that we need regardless of the changes in the tpm2-tools, since we don't support anything < 3. I've proposed that change some time ago in PR #49.

gitfineon:
Thank you for that, but the -H option was only in use from Nov'17 until Apr'18 and backwards compatibility was not implemented on purpose, so why do we care here? 
https://github.com/tpm2-software/tpm2-tools/commit/53a6e1a6465a5ff77765636423065aac7b472fd0#diff-fa8fac04c7809039a7e2b022d4841d53

I think default should be back-ported to tpm-tools version 3. Fedora already has version 3.0.4 in pipe https://www.rpmfind.net/linux/rpm2html/search.php?query=tpm2-tools and Debian/Ubuntu still uses 2.1.0 which never was compatible with clevis, right? So, why shall we wait for tpm-tools version 4?

It is broken now, and a temporary way to fix it, would be to use `--hierarchy $auth` which is valid for both.

martinezjavier:
> Thank you for that, but the -H option was only in use from Nov'17 until Apr'18 and backwards compatibility was not implemented on purpose, so why do we care here?
> tpm2-software/tpm2-tools@53a6e1a#diff-fa8fac04c7809039a7e2b022d4841d53

@gitfineon we care because the `-H` option was included as an option to some of the tpm2-tools in the 3.0.0 release. Yes, the tpm2-tools maintainer decided to break backward compatibility on the (not yet released) 4.0 version, to have a more consistent set of options across tools and the `--hierarchy` short option was one the options that wasn't used consistently in the tools.

> I think default should be back-ported to tpm-tools version 3. Fedora already has version 3.0.4 in pipe https://www.rpmfind.net/linux/rpm2html/search.php?query=tpm2-tools and Debian/Ubuntu still uses 2.1.0 which never was compatible with clevis, right? So, why shall we wait for tpm-tools version 4?

I don't think that's correct to backport `-H` to the tpm2-tools 3.x branch, since that will be a non-backward compatible change.

> It is broken now, and a temporary way to fix it, would be to use --hierarchy $auth which is valid for both.

It's not broken when using the latest released version, it doesn't work with master that's correct. But I don't think that you could expect that everything will be working with a development branch. Once tpm2-tools 4.0 is released, I think we need the patch I shared so it can work with both versions.

gitfineon:
Sorry, my assumption was your latest release [3.0.4](https://github.com/tpm2-software/tpm2-tools/releases/tag/3.0.4) from 2018-04-30 contains all commits until then, but it still uses -H option, right?

Therefore, commit https://github.com/tpm2-software/tpm2-tools/releases/tag/3.0.4 (2018-03-27) is not part of this release and your answer indicates, this change will be only part of 4.x releases.

Long option `--hierarchy` is available for both, 3.x and optional for 4.x, right? What's wrong with using this to make clevis compatible with both?

martinezjavier:
> Sorry, my assumption was your latest release 3.0.4 from 2018-04-30 contains all commits until then, but it still uses -H option, right?

@gitfineon That's correct.

> Long option --hierarchy is available for both, 3.x and optional for 4.x, right? What's wrong with using this to make clevis compatible with both?

Nothing wrong and you are right that's the best approach. Could you please propose that instead?

Unfortunately we will still need the changes for the env vars once 4.0 get released, but at least your suggestion will reduce the churn.

src/clevis-decrypt-tpm2
```diff
@@ -20,6 +20,8 @@
 
 # The owner hierarchy is the one that should be used by the Operating System.
 auth="o"
+# This hierarchy argument is default since tpm-tools version 3.0.4.
```
martinezjavier: This comment is incorrect,  in tpm2-tools 3.0.4 this argument is still required.

src/clevis-decrypt-tpm2
```diff
@@ -106,7 +108,7 @@ if ! `jose b64 dec -i- -O $TMP/jwk.priv <<< "$jwk_priv"`; then
     exit 1
 fi
 
-if ! tpm2_createprimary -Q -H "$auth" -g "$hash" -G "$key" \
+if ! tpm2_createprimary -Q $optional -g "$hash" -G "$key" \
```
martinezjavier: Just change here to `--hierarchy "$auth"`

martinezjavier: Same comments apply for the clevis-entrypt-tpm2 changes.

gitfineon: If `$optional` is empty, default hierarchy is used. Why explicitly set hierarchy, if not required?

gitfineon:
Updated commit message and environmental switch according to your proposal.
Still prefer default instead of explicit hierarchy.

src/clevis-decrypt-tpm2
martinezjavier: @gitfineon I was expecting you to only change `-H` to `--hierarchy` for now.

I don't mind if you want to squash my patch, but I don't think we should make any assumptions about tpm2-tools 4.0 until is released, just in case something else is changed in the meantime.

martinezjavier:
> I did some tests and noticed also some context flags have changed (short and long version) since

@gitfineon right, that's a recent change. That's why I think we need to wait until things settle down for the upcoming tpm2-tools release since there are a lot of changes to make it the options consistent. Hopefully after this major release this won't change anymore.

> Are you using the branches 1.x for tpm2-tss and 3.x for tpm2-tools? I had an issue to decrypt if

That's correct, I'm using the Fedora 28 packages that contains the latest released versions: tpm2-tss 1.4.0 and tpm2-tools 3.0.4.

> worked fine with defined PCRs. Maybe you can reproduce the issue

I'm not able to reproduce the issue. Maybe you can start with a clean build just to make sure that are not left overs from previous builds.

martinezjavier:
This PR has been superseded by PR #114 so it can be closed now.

PR #57, opened 2018-05-28
Allow to use clevis without dracut

gitfineon:
Automatic decryption can be used for removable devices without having
an encrypted rootfs. Some systems still use other tools to generate
their initramfs and the package requirement to have dracut installed
blocks configure if you want to use clevis without dracut.

Signed-off-by: Michael Brandl <git@fineon.pw>

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -5,25 +5,29 @@ To build and install the Clevis software the following software packages
 are required. In many cases dependencies are platform specific and so the
 following sections describe them for the supported platforms.
 
-## Linux:
-* Autoconf
-* Autoconf archive
-* Automake
-* Libtool
-* C compiler
-* C Library Development Libraries and Header Files
-* [jose](https://github.com/latchset/jose)
-* [luksmeta](https://github.com/latchset/luksmeta)
-* [audit-libs](https://github.com/linux-audit/audit-userspace)
-* [udisks2](https://github.com/storaged-project/udisks)
-* [OpenSSL](https://github.com/openssl/openssl)
-* [desktop-file-utils](https://cgit.freedesktop.org/xdg/desktop-file-utils)
-* [pkg-config](https://cgit.freedesktop.org/pkg-config)
-* [systemd](https://github.com/systemd)
-* [dracut](https://github.com/dracutdevs/dracut)
-* [tang](https://github.com/latchset/tang)
-* [curl](https://github.com/curl/curl)
-* [tpm2-tools](https://github.com/tpm2-software/tpm2-tools)
+## Linux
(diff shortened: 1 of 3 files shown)


gitfineon:
Just saw there is an open PR https://github.com/latchset/clevis/pull/35 to add initramfs-tools, but i still think it would be great to consider the use case without initramfs modification.

This patch should not break anything and is a interim solution.

gitfineon:
Hey @npmccallum, @blaufish, @martinezjavier,
I think clevis can also be useful without changing initramfs, do you agree?

npmccallum:
@gitfineon Yes, I agree. However, this patch also needs to conditionally install the dracut bits.

npmccallum:
@blaufish Yes, I will definitely merge a complete patch to make dracut optional.

gitfineon:
Hey, thanks for the feedback! 
I'm not experienced with autoconf, but tried to make it conditional and added hints to INSTALL.md

gitfineon:
Updated INSTALL.md concerning dependencies. Splitted into general, PIN specific and initramfs and added short hint, why clevis requires them.

INSTALL.md
```diff
@@ -5,25 +5,29 @@ To build and install the Clevis software the following software packages
...
+* [LUKSMeta](https://github.com/latchset/luksmeta) (Adds metadata to LUKS volume)
+* [linux-audit](https://github.com/linux-audit/audit-userspace) (Userspace lib of Linux Auditing Framework)
+* [udisks2](https://github.com/storaged-project/udisks) (Storage management deamon, tools and libs)
+* [desktop-file-utils](https://cgit.freedesktop.org/xdg/desktop-file-utils) (Linux .desktop files / autostart)
+* [pkg-config](https://cgit.freedesktop.org/pkg-config) (Dependency discovery with Autotools)
+* [systemd](https://github.com/systemd) (Control clevis akspass service)
+
+PIN specific dependencies
+* [tang](https://github.com/latchset/tang) (required for pin **Tang**)
+* [curl](https://github.com/curl/curl) (required for pin **HTTP**)
```
npmccallum: curl is required for both Tang and HTTP

gitfineon: done

mhalano:
@gitfineon Any news about that?

gitfineon:
> @gitfineon Any news about that?

I was able to run it without dracut on a debian platform for a partition other than the root partition.
Just printing a warning, if dracut is not in use. This is what my pull request solved.

@npmccallum did not merge it. Tell me if there is still interest for this intermediate step.

dkopecek:
Unfortunately, this is now obsolete as the project migrated to meson build system.

PR #65, opened 2018-07-12
Use asciidoc as the manual page source format

dkopecek:
To generate the manual page, you can use the a2x utility
from the asciidoc package:

 $ a2x -v -f manpage clevis.1.adoc
 ...

or use the provided doc/Makefile:

 $ cd doc/
 $ make


dkopecek:
I've added a separate commit which implements the automatic manual page generation.

doc/clevis-decrypt.1.adoc
```diff
@@ -0,0 +1,21 @@
+CLEVIS-DECRYPT(1)
+=================
+:doctype: manpage
+
+== NAME
+
+clevis-decrypt - Decrypts using the policy defined at encryption time
```
npmccallum: Please use spaces instead of dashes.

dkopecek: I think it's a convention to use dashes in the NAME section as it refers to the manual page name. See for example `man git apply`. The SYNOPSIS section on the other hand refers to the usage of the actual command.

npmccallum: Thanks!

PR #67, opened 2018-07-16
Globbing for executable matches in main clevis

jvymazal:
When glob fails we can now emit more detailed and
"intelligent" error message stating explicitly which
part of input line is causing problems. Also, this relaxes
requirement for dashed arguments for clevis-* scripts.


jvymazal:
@npmccallum could you please review this?

PR #77, opened 2018-10-18
Update clevis-luks-askpass nc command line change for ubuntu compatibility

michaelbarkdoll:
Added additional compatibility check for nc and socat.

This is in reference to:
https://github.com/latchset/clevis/pull/51
https://github.com/latchset/clevis/pull/50

This is best practice for bash scripts according to:
https://stackoverflow.com/questions/592620/how-to-check-if-a-program-exists-from-a-bash-script


michaelbarkdoll-siucs:
This appears to have been fixed by a debian configure patch.

michaelbarkdoll:
This issue was fixed by a debian configure patch.

PR #96, opened 2019-03-31
All kinds of fixes

maage:
As this touches almost all parts it is not thoroughly tested. At least tpm2 luks, bind, unbind, dracut boot should work.

There is some security implications as many strings are not quoted, so malicious paths etc can cause havoc. And shell commands might expand \n etc in passwords or others strings.

I hope I was able to split patches enough.


src/luks/clevis-luks-bind
```diff
@@ -120,16 +120,16 @@ fi
...
+ *) old="$(/bin/cat "$KEY")";;
 esac
 
 # Add the new key
 if [ -n "$SLT" ]; then
     cryptsetup luksAddKey --key-slot "$SLT" "$DEV"
 else
-    SLT=`cryptsetup luksAddKey -v "$DEV"
-        | sed -rn 's|^Key slot ([0-9]+) created\.$|\1|p'`
+    SLT="$(cryptsetup luksAddKey -v "$DEV"
```
ZelenyMartin: It says:
/usr/bin/clevis-luks-bind: command substitution: line 135: syntax error near unexpected token `|'
Joint those two lines to one.

ZelenyMartin: Hi, we will be able to merge the requests. No other issues were found. Can you just fix this BASH syntax error with pipe on a new line? Thanks.

dkopecek:
Closing this PR as the author is not responding. Let's continue in #104.

PR #113, opened 2019-08-16
Added regen and report modules for clevis

sergio-correia:
* Usage:
clevis luks report -d DEVICE -s SLOT
  - reports rotation of keys
clevis luks regen -d DEVICE -s SLOT
  - regenerate actual LUKS metadata and change the key for cryptsetup

Built upon work of radosroka in PR#69.


sergio-correia:
I will close this, rebase and repost it in two parts, one for regen and another for report, so that it's simpler to review.

PR #118, opened 2019-09-04
Speed up PBKDF

pekon:
From cryptsetup docs:
" A PBKDF is used for increasing dictionary and brute-force attack cost for keyslot passwords."

Rationale: $key and LUKS Master Key have the same entropy. Therefore the cost of attack on each is about the same and we do not need to waste time in PBKDF. This speeds up mount times by ~2 seconds.

### DIFF
--- a/src/luks/clevis-luks-bind
+++ b/src/luks/clevis-luks-bind
@@ -127,9 +127,9 @@ esac
 
 # Add the new key
 if [ -n "$SLT" ]; then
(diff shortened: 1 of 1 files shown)


sarroutbi:
This is override by #339. May I close it?

sarroutbi:
Closing PR due to solution provided by #339. Please, feel free to re-open it if you consider necessary

PR #119, opened 2019-09-17
pins/tang: check that key derivation key is available

sergio-correia:
Sometimes, tang may not generate an advertisement with 'deriveKey' [1],
so let's check whether we have the key derivation key before proceeding
to encryption.

Original patch by D. Kopecek.

[1] Related to https://github.com/latchset/tang/issues/23

### DIFF
--- a/src/pins/tang/clevis-encrypt-tang
+++ b/src/pins/tang/clevis-encrypt-tang
@@ -115,7 +115,11 @@ elif [ "$thp" != "any" ] && \
 fi
 
 ### Perform encryption
-enc="$(jose jwk use -i- -r -u deriveKey -o- <<< "$jwks")"
+if ! enc="$(jose jwk use -i- -r -u deriveKey -o- <<< "$jwks")"; then
+    echo "Key derivation key not available!" >&2
+    exit 1
+fi
+
 jose fmt -j "$enc" -Og keys -A || enc="{\"keys\":[$enc]}"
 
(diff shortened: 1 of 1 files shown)


QC-FreAk:
I've found when the derivation key is not available, there are ownership issues of the key files in tang. I'm using RHEL 7.6, by default files in /var/db/tangd are owned by root:tang; this always seems to present the missing derivation key issue in clevis. If I change their ownership to tang:tang and restart tangd.socket, clevis binds work flawlessly.

sergio-correia:
@QC-FreAk: you can also try `systemctl restart tangd-update` when it happens. What seem to happen is that the keys are generated on startup by `tangd-keygen`, but `tangd-update` runs before the keygen step completes, and at that point, we only generated the first of the two keys.

You can reproduce it artificially by doing the following:
1) stop tangd.socket
2) remove the keys in `/var/db/tang` (backup them first, if they are actually used for something, naturally)
3) create the first key: `jose jwk gen -i '{"alg":"ES512"}' > /var/db/tang/sig.jwk`
4) start tangd.socket
5) try to encrypt something: `clevis encrypt tang '{"url":"localhost"}' <<< "test"`

And you should get the message about the key derivation key not being available.

By the way, `tangd-keygen` did not run in step 4) above because there was a key already (`sig.jwk` from step `3)`). And you currently have a broken setup. You can create the missing key manually with `jose jwk gen -i '{"alg":"ECMR"}'  > /var/db/tang/exc.sig` and restart `tangd.socket` (or restart `tangd-update.service`).

PR #132, opened 2019-11-24
Add Travis-CI integration

sergio-correia:


sergio-correia:
@npmccallum: could you take a look into this PR, please? I added some tests earlier (#134) and now I am adding Travis-CI.

Build is currently failing due to the following two tests: 
1. `unbind-luks1` (which should be fixed by #130), and 
2. `bind-luks1`, which fails if using old cryptsetup (< 2.0.4), since we don't have consistent messages about slots being added/removing and our parsing fails. We will be addressing this probably in (#135), still ongoing.

.travis/install
```diff
@@ -0,0 +1,56 @@
...
+        libudisks2-dev libaudit-dev; do
+
+        sleep 5
+    done
+    ;;
+
+fedora:*)
+    dnf -y clean all
+    dnf -y --setopt=deltarpm=0 update
+    dnf -y install ${COMMON} pkgconfig openssl-devel jansson-devel \
```
martinezjavier: Can we just use dnf builddep clevis here instead? And same for Debian/Ubuntu. Otherwise the build dependency list will have to be maintained.

sergio-correia:
> Looks good to me. I had just a small comment but it can even be done as a follow-up if you agree.

Thanks for the review! I will merge this for now and apply your suggestion for improving the maintenance of the build dependencies sometime later.

PR #134, opened 2019-11-24
Add tests for LUKS binding and unbinding

sergio-correia:
Right now a couple of tests that includes binding fail when running
in a system with old (< 2.0.4) cryptsetup.

Once we have travis-ci integrated, it will be simpler to make sure
common cases are working.


sergio-correia:
> Test expectedly fails on wrong current state of master branch code.

Thanks for reviewing.

Yes, we currently have 2 tests failing:
`unbind-luks1` - this issue is addressed in #130 
`bind-luks1` - this one fails when using cryptsetup < 2.0.4, and we will address it in #135

PR #138, opened 2019-11-28
Add clevis luks list command

sergio-correia:
Reviving Javier's idea in PR#24, we now have a command to list the pins
bound to a LUKS volume.

Usage:
clevis luks list -d DEV [-s SLT]

Examples:

clevis luks list -d device
1: sss '{"t":1,"pins":{"tang":[{"url":"addr1"},{"url":"addr2"}],"tpm2":[{"hash":"sha256","key":"ecc"}],"sss":{"t":1,"pins":{"tang":[{"url":"addr3"}]}}}}'
2: tang '{"url":"addr"}'
3: tpm2 '{"hash":"sha256","key":"ecc","pcr_bank":"sha1","pcr_ids":"7"}'

clevis luks list -d device -s 3
3: tpm2 '{"hash":"sha256","key":"ecc","pcr_bank":"sha1","pcr_ids":"7"}'

### DIFF
--- a/src/luks/clevis-luks-common-functions
+++ b/src/luks/clevis-luks-common-functions
@@ -1,0 +1,283 @@
+#!/bin/bash -e
+# vim: set tabstop=8 shiftwidth=4 softtabstop=4 expandtab smarttab colorcolumn=80:
+#
+# Copyright (c) 2019 Red Hat, Inc.
+# Author: Sergio Correia <scorreia@redhat.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+# valid_slot() will check whether a given slot is possibly valid, i.e., if it
+# is a numeric value within the specified range.
+valid_slot() {
+    local SLT="${1}"
+    local MAX_SLOTS="${2}"
+    case "${SLT}" in
+        ''|*[!0-9]*)
+            return 1
+            ;;
+        *)
+            # We got an integer, now let's make sure it is within the
+            # supported range.
+            if [ "${SLT}" -ge "${MAX_SLOTS}" ]; then
+                return 1
+            fi
+            ;;
+    esac
+}
+
+# clevis_luks_read_slot() will read a particular slot of a given device, which
+# should be either LUKS1 or LUKS2. Returns 1 in case of failure; 0 in case of
+# success.
+clevis_luks_read_slot() {
+    local DEV="${1}"
+    local SLT="${2}"
+
+    if [ -z "${DEV}" ] || [ -z "${SLT}" ]; then
+        echo "Need both a device and a slot as arguments." >&2
+        return 1
+    fi
+
+    local DATA_CODED=''
+    local MAX_LUKS1_SLOTS=8
+    local MAX_LUKS2_SLOTS=32
+    if cryptsetup isLuks --type luks1 "${DEV}"; then
+        if ! valid_slot "${SLT}" "${MAX_LUKS1_SLOTS}"; then
+            echo "Please, provide a valid key slot number; 0-7 for LUKS1" >&2
+            return 1
+        fi
+
+        if ! luksmeta test -d "${DEV}"; then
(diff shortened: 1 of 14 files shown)


sergio-correia:
@martinezjavier: could you review this one, please? I have rebased it on top of master

martinezjavier:
> @martinezjavier: could you review this one, please? I have rebased it on top of master

Sure, I'll do it tomorrow. Great that you brought this back.

src/luks/clevis-luks-common-functions
```diff
@@ -0,0 +1,283 @@
...
+        if ! valid_slot "${SLT}" "${MAX_LUKS1_SLOTS}"; then
+            echo "Please, provide a valid key slot number; 0-7 for LUKS1" >&2
+            return 1
+        fi
+
+        if ! luksmeta test -d "${DEV}"; then
+            echo "The ${DEV} device is not valid!" >&2
+            return 1
+        fi
+
```
martinezjavier, line 65: I think is more natural to first test if the device is LUKS and then test if the slot number is between the supported range. Although in practice it doesn't matter.

src/luks/clevis-luks-common-functions
```diff
@@ -0,0 +1,283 @@
...
+# clevis_luks_print_pin_config() will print the config of a given pin; i.e.
+# for tang it will display the associated url address, and for tpm2, the
+# properties in place, like the hash, for instance.
+clevis_luks_print_pin_config() {
+    local P="${1}"
+    local decoded="${2}"
+
+    local content
+    if ! content="$(jose fmt -j- -g clevis -g "${P}" -o- <<< "${decoded}")" \
+                    || [[ -z "${content}" ]]; then
```
martinezjavier, line 152: The script uses both bash specific `[[` and `[`, I think that either is OK but that should stick with one and use it consistently.

src/luks/clevis-luks-common-functions
```diff
@@ -0,0 +1,283 @@
...
+        return 1
+    fi
+
+    local pin=
+    case "${P}" in
+    tang)
+        local url
+        url="$(jose fmt -j- -g url -u- <<< "${content}")"
+        pin=$(printf '{"url":"%s"}' "${url}")
+        printf "tang '%s'" "${pin}"
```
martinezjavier, line 162: The `tang` ping also has other properties like `thp` and `adv`. Do you think that users may also be interested in printing those? (I'm not that familiar with tang to answer this).

src/luks/clevis-luks-common-functions
```diff
@@ -0,0 +1,283 @@
...
+}
+
+# clevis_luks_process_sss_pin() will receive a JWE with information on the sss
+# pin config, and also its associated threshold, and will extract the info.
+clevis_luks_process_sss_pin() {
+    local jwe="${1}"
+    local threshold="${2}"
+
+    local sss_tang
+    local sss_tpm2
```
martinezjavier, line 223: I think this function can be mode more simpler and these two variables be removed...

src/luks/clevis-luks-common-functions
```diff
@@ -0,0 +1,283 @@
...
+
+    local coded
+    for coded in $(jose fmt -j- -Og jwe -Af- <<< "${jwe}"| tr -d '"'); do
+        if ! pin_cfg="$(clevis_luks_decode_pin_config "${coded}")"; then
+            continue
+        fi
+        read -r pin cfg <<< "${pin_cfg}"
+        case "${pin}" in
+        tang)
+            sss_tang="${sss_tang},${cfg}"
```
martinezjavier, line 237: ... having something like the folowing:

```
sss_tang="${sss_tang},${cfg}"
cfg=$(clevis_luks_join_sss_cfg "tang" "${sss_tang}")
```

martinezjavier:
@sergio-correia sorry for the delay, I've finally reviewed this. It looks good to me! I just had two small comments.

Since is not trivial shell code I had to read it twice, I also wrote some comments but then realized that it was just me misunderstanding the code, so I deleted them.

src/luks/clevis-luks-common-functions
sergio-correia: Good point. For `tang`, besides the `url` we store the `adv`, as an array of keys, something like:

```
"adv": {
        "keys": [
          {
            "alg": "ECMR",
            <snip>
          },
          {
            "alg": "ES512",
            <snip>
          }
        ]
      }
```

I am not sure users would be interested in printing this. The way it is now, we output a valid configuration that can be reused to rebind the slot, so I think we could keep printing just the `url`. Please, let me know your thoughts.

sergio-correia:
> @sergio-correia sorry for the delay, I've finally reviewed this. It looks good to me! I just had two small comments.
> 
> Since is not trivial shell code I had to read it twice, I also wrote some comments but then realized that it was just me misunderstanding the code, so I deleted them.

Thanks for the review! I will incorporate your suggestions and update it.

src/luks/clevis-luks-common-functions
sergio-correia: I am now sticking with `[`.

PR #140, opened 2019-11-28
Validate the pcr_ids and error out if parsing fails

sergio-correia:
As of now, our parsing of pcr_ids fails if we try to use an actual JSON
array, and when it happens, we end up silently using no policy at all,
which is certainly not what the user requested/intended.

This patch improves our parsing to support reading both strings and JSON
arrays for the pcr_ids propery. If we encounter an error, we report it
and exit.

Fixes: #103

### DIFF
--- a/src/pins/tpm2/clevis-encrypt-tpm2
+++ b/src/pins/tpm2/clevis-encrypt-tpm2
@@ -103,7 +103,29 @@ key="$(jose fmt -j- -Og key -u- <<< "$cfg")" || key="ecc"
 
 pcr_bank="$(jose fmt -j- -Og pcr_bank -u- <<< "$cfg")" || pcr_bank="sha1"
 
-pcr_ids="$(jose fmt -j- -Og pcr_ids -u- <<< "$cfg")" || true
+# Issue #103: We support passing pcr_ids using both a single string, as in
+# "1,3", as well as an actual JSON array, such as ["1,"3"]. Let's handle both
+# cases here.
+if [[ ${cfg// /} != '{}' ]] \
+    && ! pcr_ids="$(jose fmt -j- -Og pcr_ids -u- 2>/dev/null <<< "$cfg")"; then
+
+    # We failed to parse a string, so let's try to parse a JSON array instead.
+    if jose fmt -j- -Og pcr_ids -A 2>/dev/null <<< "${cfg}"; then
+        # OK, it is an array, so let's get the items and form a string.
+        pcr_ids=
+        for pcr in $(jose fmt -j- -Og pcr_ids -Af- <<< "${cfg}" \
+                     | tr -d '"'); do
+            pcr_ids=$(printf '%s,%s' "${pcr_ids}" "${pcr}")
+        done
+        # Now let's remove the leading comma.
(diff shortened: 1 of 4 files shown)


sergio-correia:
@martinezjavier: Could you take a look into this PR, please?

martinezjavier:
> @martinezjavier: Could you take a look into this PR, please?

@sergio-correia Sorry, I was too busy at the end of the year and then on holidays. I'll take a look to this tomorrow if you don't mind

sergio-correia:
> > @martinezjavier: Could you take a look into this PR, please?
> 
> @sergio-correia Sorry, I was too busy at the end of the year and then on holidays. I'll take a look to this tomorrow if you don't mind

No problem, and thanks in advance!

src/pins/tpm2/clevis-encrypt-tpm2
```diff
@@ -103,7 +103,29 @@ key="$(jose fmt -j- -Og key -u- <<< "$cfg")" || key="ecc"
...
+if [[ ${cfg// /} != '{}' ]] \
+    && ! pcr_ids="$(jose fmt -j- -Og pcr_ids -u- 2>/dev/null <<< "$cfg")"; then
+
+    # We failed to parse a string, so let's try to parse a JSON array instead.
+    if jose fmt -j- -Og pcr_ids -A 2>/dev/null <<< "${cfg}"; then
+        # OK, it is an array, so let's get the items and form a string.
+        pcr_ids=
+        for pid in $(jose fmt -j- -Og pcr_ids -Af- <<< "${cfg}" \
+                     | tr -d '"'); do
+            pcr_ids=$(printf '%s,%s' "${pcr_ids}" "${pid}")
```
martinezjavier: I think that most people associate pid with Process ID so I probably will call this variable pcr or just id instead to make the code less confusing to people not familiar with it.

But feel free to ignore my comment if you disagree.

martinezjavier:
@sergio-correia first of all, sorry for the late review. I had to re-read all the issues and PR to remember what this was about.

I thought that this was a bug in jose, but after reading your comment https://github.com/latchset/jose/issues/68#issuecomment-560111999 I agree with you and understand now that's not the case.

The changes looks good to me, I just had a small cosmetic suggestion. Thanks a lot for fixing this issue!

src/pins/tpm2/clevis-encrypt-tpm2
sergio-correia: I agree; I associate it with Process ID myself as well, at a first sight; I will use `pcr` instead, as you suggest.

sergio-correia:
> @sergio-correia first of all, sorry for the late review. I had to re-read all the issues and PR to remember what this was about.
> 
> I thought that this was a bug in jose, but after reading your comment [latchset/jose#68 (comment)](https://github.com/latchset/jose/issues/68#issuecomment-560111999) I agree with you and understand now that's not the case.
> 
> The changes looks good to me, I just had a small cosmetic suggestion. Thanks a lot for fixing this issue!

Thanks also for your review. I will do the change you suggested and push the fix.

PR #144, opened 2019-12-02
Search for tang in libdir used by Ubuntu 18.04 packages

### LINKED ISSUE #143: Clevis doesn't find tang on Ubuntu 18.04
The path is set to `libexecdir`, which is `/usr/libexec`, but [on Ubuntu these files are installed under `/usr/lib/x86_64-linux-gnu`](https://packages.ubuntu.com/bionic/amd64/tang/filelist).

Workaround:

```
sudo ln -s /usr/lib/x86_64-linux-gnu/tangd /usr/libexec/
sudo ln -s /usr/lib/x86_64-linux-gnu/tangd-keygen /usr/libexec/
sudo ln -s /usr/lib/x86_64-linux-gnu/tangd-update /usr/libexec/
```

bviktor:
Fixes #143

### DIFF
--- a/src/pins/tang/meson.build
+++ b/src/pins/tang/meson.build
@@ -6,16 +6,19 @@ actv = find_program(
 kgen = find_program(
   join_paths(libexecdir, 'tangd-keygen'),
   '/usr/libexec/tangd-keygen',
+  '/usr/lib/x86_64-linux-gnu/tangd-keygen',
   required: false
 )
 updt = find_program(
   join_paths(libexecdir, 'tangd-update'),
   '/usr/libexec/tangd-update',
+  '/usr/lib/x86_64-linux-gnu/tangd-update',
   required: false
 )
 tang = find_program(
   join_paths(libexecdir, 'tangd'),
   '/usr/libexec/tangd',
+  '/usr/lib/x86_64-linux-gnu/tangd',
   required: false
 )
 curl = find_program('curl', required: false)
@@ -32,6 +35,7 @@ if curl.found()
       join_paths(meson.source_root(), 'src'),
       meson.current_source_dir(),
       '/usr/libexec',
+      '/usr/lib/x86_64-linux-gnu',
       libexecdir,
       separator: ':'
     )


sergio-correia:
I added this PR in a branch that gets built by the CI and it seems this is not enough to find tang in Ubuntu 18.04: https://travis-ci.org/sergio-correia/clevis/jobs/620080959#L1657

It looks like it's now looking for it in `/usr/local/lib/x86_64-linux-gnu/`, since the default prefix is `/usr/local`, however it gets installed by default in the `/usr` prefix, so you need to look for the tang binaries there as well. 

Perhaps something like the following in the `find_program()`paths would do: 
`join_paths('/usr', get_option('libdir'), 'tangd')`

bviktor:
Well I didn't want to bring it up, but the whole thing is messed up, because the path it looks for the dependencies in is equivalent to the path you configure clevis for, which doesn't make much sense.

Not to mention the initramfs-tools unlocker, which always looks for `/usr/bin/tpm*` during `update-initramfs`, even if you built using the `/usr/local` prefix.

Currently it fails with _any_ prefix, with this patch it at least works for the `/usr`/ prefix. Assuming all your dependencies are there as well (tpm2-tools, etc.).

I'm not sure how to fix the problem you mentioned, but that requires way more changes, coz like I said, it's already the way it is.

bviktor:
In any case, [Meson / Built-in options](https://mesonbuild.com/Builtin-options.html) says

> prefix defaults to C:/ on Windows, and /usr/local/ otherwise. **You should always override this value.**

Naturally, I always build clevis with `--prefix=/usr` to avoid these problems, that are present even without this patch.

It doesn't seem trivial to look both in `/usr` and `/usr/local`, and for every dependency. Maybe `pkg-config` can help, but that relies on `.pc` files. For instance, 18.04 dracut doesn't have one, but 18.10 does. Not that dracut is too relevant on Ubuntu, but it's a good example for what might happen if clevis starts relying on pkg-config too much.

But it might still be better than searching for cherry-picked binaries in various folders. I'm really not sure, as I'm not too familiar with meson/ninja at all.

bviktor:
How do you feel about hardcoding /usr/lib/x86_64-linux-gnu/tang*, much like it's already done for Fedora with /usr/libexec/tang*?

sergio-correia:
> How do you feel about hardcoding `/usr/lib/x86_64-linux-gnu/tang*`, much like it's already done for Fedora with `/usr/libexec/tang*`? This path really only applies to the package version, and that never changes. Even the 19.10 and the upcoming 20.04 package uses this path.

It does not look pretty, but we already have something like this, so I think it's OK. Please update the PR and I will have travis to test it again.

bviktor:
Thank you. I updated the commit accordingly :)

bviktor:
Awesome, thank you!

PR #152, opened 2019-12-12
initramfs: Add support for LUK2

superm1:
I did testing with master on Ubuntu 20.04 daily images and found some lacking aspects for LUKS2.  I ported some stuff over from the systemd script to work in the initramfs-tools one.

I confirmed this works properly with FDE on Ubuntu 20.04.

### DIFF
--- a/src/initramfs-tools/scripts/local-top/clevis.in
+++ b/src/initramfs-tools/scripts/local-top/clevis.in
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash -e
 #
 # Copyright (c) 2017 Red Hat, Inc.
 # Copyright (c) 2017 Shawn Rose
@@ -51,6 +51,31 @@ get_askpass_pid() {
     done
 }
 
+luks1_decrypt() {
+    luksmeta load "$@" \
+        | clevis decrypt
+
+    local rc
+    for rc in "${PIPESTATUS[@]}"; do
+        [ $rc -eq 0 ] || return $rc
+    done
+    return 0
+}
+
+luks2_jwe() {
+    # jose jwe fmt -c outputs extra \n, so clean it up
+    cryptsetup token export "$@" \
+        | jose fmt -j- -Og jwe -o- \
+        | jose jwe fmt -i- -c \
+        | tr -d '\n'
+
+    local rc
+    for rc in "${PIPESTATUS[@]}"; do
+        [ $rc -eq 0 ] || return $rc
+    done
+    return 0
+}
+
 # Wait for askpass, and then try and decrypt immediately. Just in case
 # there are multiple devices that need decrypting, this will loop
 # infinitely (The local-bottom script will kill this after decryption)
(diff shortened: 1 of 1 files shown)


superm1:
^ @therealjuanmartinez

superm1:
@sergio-correia any updates here?

sergio-correia:
> @sergio-correia any updates here?

Sorry for the delay. It looks good to me, but I was going to test it before merging. I will try to get it done today or tomorrow.

sergio-correia:
@superm1: I tested this on a laptop and a VM and had issues in both cases; while the issues fall out of the scope of this PR itself, which is merely adding support for LUKS2, you may want to investigate them better:

1) on the laptop, I was using a usb network card which was not detected, hence no network at boot time. I could see the "Error communicating with the server" message displayed by `clevis-decrypt-tang`, which was expected, since there was no network available.

2) on the VM, I saw the DHCP request and that there was network available, however no indications from `clevis-decrypt-tang`. I was just being asked the passphrase to unlock the disk. After some debugging, I noticed the `configure_networking` call (`local-top/clevis.in`) was blocking; once I added `&` to run it in the background, I was able to get it to work.

superm1:
> @superm1: I tested this on a laptop and a VM and had issues in both cases; while the issues fall out of the scope of this PR itself, which is merely adding support for LUKS2, you may want to investigate them better:

So I did my testing using TPM2 (which is the more interesting case to me since TPMs are available very widely but a tang server requires some setup in advance).

I hence didn't hit any networking related issues in my testing.  If you would be able to do your laptop test with the TPM2 I think you'd have similar success.

>on the laptop, I was using a usb network card which was not detected, hence no network at boot time. I could see the "Error communicating with the server" message displayed by clevis-decrypt-tang, which was expected, since there was no network available.

Presumably this network card is functional via some kernel module in the rootfs, just missing from initramfs, right?  I think you may try to add it explicitly to `/etc/initramfs/modules` and rebuild your initramfs.  It should hopefully take effect then.

We might have a bug somewhere initramfs-tools that it didn't pull it on it's own though.  Which module is it?

> on the VM, I saw the DHCP request and that there was network available, however no indications from clevis-decrypt-tang. I was just being asked the passphrase to unlock the disk. After some debugging, I noticed the configure_networking call (local-top/clevis.in) was blocking; once I added & to run it in the background, I was able to get it to work.

That's a bit odd.  `configure_networking` shouldn't be blocking.  Could you add a `set -x` into `scripts/functions` where it's defined?  Maybe we can see in the logging what it's actually blocking on.

> I was able to get it to work.

As in successful test with LUKS2?  :)

sergio-correia:
> So I did my testing using TPM2 (which is the more interesting case to me since TPMs are available very widely but a tang server requires some setup in advance).
> 
> I hence didn't hit any networking related issues in my testing. If you would be able to do your laptop test with the TPM2 I think you'd have similar success.
> 

tpm2 worked here as well, once I had `--prefix=/usr`, so that it could find `libtss2-tcti-device.so` to include in the initramfs.

> > on the laptop, I was using a usb network card which was not detected, hence no network at boot time. I could see the "Error communicating with the server" message displayed by clevis-decrypt-tang, which was expected, since there was no network available.
> 
> Presumably this network card is functional via some kernel module in the rootfs, just missing from initramfs, right? I think you may try to add it explicitly to `/etc/initramfs/modules` and rebuild your initramfs. It should hopefully take effect then.
> 
> We might have a bug somewhere initramfs-tools that it didn't pull it on it's own though. Which module is it?
> 

Yeah, something like this should work, however I tried quickly here and it still did not; I added `usbnet` and `r8152` to `/etc/initramfs-tools/modules` and rebuilt the initramfs (and confirmed they were added).

> > on the VM, I saw the DHCP request and that there was network available, however no indications from clevis-decrypt-tang. I was just being asked the passphrase to unlock the disk. After some debugging, I noticed the configure_networking call (local-top/clevis.in) was blocking; once I added & to run it in the background, I was able to get it to work.
> 
> That's a bit odd. `configure_networking` shouldn't be blocking. Could you add a `set -x` into `scripts/functions` where it's defined? Maybe we can see in the logging what it's actually blocking on.
> 
> > I was able to get it to work.
> 
> As in successful test with LUKS2? :)

Yep. LUKS2 and now tested working with both tang and tpm2 pins. I will merge this PR, thanks!

PR #166, opened 2020-02-16
tpm2: fix pcr_ids validation

sergio-correia:
In PR#140 we improved the validation for pcr_ids, however we introduced
an error in this validation, which broke cases in which we specify
something in the configuration, but not the pcr_ids attribute.

E.g.: '{"key": "ecc"}' now fails with "Parsing the requested policy
failed!".

In this commit we fix this issue and add a new test case to cover it.


sergio-correia:
@martinezjavier: could you take a look into this one, please?

sergio-correia:
Thanks! Yeah, it escaped mine as well. Looking at the bright side, our coverage is improving.

PR #182, opened 2020-04-04
Check if network interfaces have a link in short intervals instead of…

yuergen:
… waiting a fixed amount of time.

This should help improving some if the issues with the initialization of the network devices discussed in #145. 

I have tested this inside a virtual machine using the current beta of focal/Ubuntu 20.04. The modifications are working.


sergio-correia:
Thanks for this. I merged a different approach in f670383 (initramfs: Wait for interface to appear instead of checking for carriers).

PR #187, opened 2020-04-24
tpm2: fix parsing when there are spaces in the list of PCR IDs

sergio-correia:
Configurations like '"pcr_ids": "0, 1"}' were failing because we
did not properly parse the PCR IDs.

Quick test:
echo foo | clevis encrypt tpm2 '{"pcr_ids": "0, 1"}' | clevis decrypt
Unsealing jwk from TPM failed!


sergio-correia:
@martinezjavier: would you take a look into this one, please?

sergio-correia:
> Indeed, the fix looks good to me.

This was fast, many thanks!

PR #189, opened 2020-05-02
initramfs-tools: Copy a certificate bundle so curl can do https

g2p:
The path is the one used on Debian and Ubuntu, which should match
where initramfs-tools is used.


g2p:
Would you apply this?  This is still required when tang uses https.

sergio-correia (approved):

sergio-correia:
Thanks, looks good to me.

gb-123-git:
Is this patch still required ?

g2p:
> Is this patch still required ?

It is.  At least on systems that use initramfs-tools.

gb-123-git:
Any plans to merge this ?

sarroutbi (approved):
LGTM

PR #192, opened 2020-05-16
Use clevis-luks-askpass systemd units in dracut

### LINKED ISSUE #150: Slow unlock and passphrase request remains during boot
OS: Fedora 31
RPMs: clevis-11-10.fc31.x86_64 manually installed
(The issue is the same as 11-8 from dnf)
After a normal installation `sudo clevis luks bind tpm2 ...` and `sudo dracut -f` unlock happens automatically as expected. But before unlocking happens it take almost 2-3s (On an i7 - NVMe SSD, so it's not performance issue). Also after disk unlock, passphrase request isn't dismissed and remains during boot or upgrade reboot (During updates install remains the passphrase request and under is written the update progress).
One time, I don't know how, I got it to unlock almost istantly and with passphrase request dismissal.

ArturGaspar:
With `Before=systemd-ask-password-console.service systemd-ask-password-plymouth.service` it also fixes #150, so that the user only gets prompted for a password if the clevis unlocker fails.

I am using this with the tpm2 pin and it works correctly both in case of success or failure. If there is interest in merging this I can also test if it works with the tang pin and others.

### DIFF
--- a/src/luks/systemd/clevis-luks-askpass.path
+++ b/src/luks/systemd/clevis-luks-askpass.path
@@ -1,7 +1,8 @@
 [Unit]
 Description=Clevis systemd-ask-password Watcher
-Before=remote-fs-pre.target
-Wants=remote-fs-pre.target
+DefaultDependencies=no
+Conflicts=shutdown.target
+Before=basic.target shutdown.target remote-fs-pre.target
 
 [Path]
 PathChanged=/run/systemd/ask-password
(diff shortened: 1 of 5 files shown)


sergio-correia:
Thanks for this. And apologies for not getting to it earlier. We merged something similar in cbb64c4 (dracut: favour systemd units over dracut hooks).

I filed this other issue with plymouth regarding the passphrase not being dismissed when a LUKS device is unlocked non-interactively, in our case, with clevis: https://gitlab.freedesktop.org/plymouth/plymouth/-/issues/126 (plymouth splash is not dismissed when LUKS device is unlocked non-interactively) and it would help #150 once resolved.

PR #203, opened 2020-05-24
pin template with documentation, sample "file" pin

cbiedl:
At least for me, writings pins was hard since I had to figure out the concept behind them, and when studying the existing ones, finding out what's generic pin logic and what it specific for that pin wasn't that easy either.

That should not be repeated for anyone else, so I created `src/pins/template/` and filled it my several files and a lot of explanations embedded.

Also a new pin "file" as a demonstration of a minimal pin that uses a file to store the key. I'd consider this suitable for educational puposes only, but you think it's worth to include it in the official distribution, go ahead. Actually, this was im preparation of #185 which will get a fix soon.

### DIFF
--- a/src/pins/file/clevis-encrypt-file.1.adoc
+++ b/src/pins/file/clevis-encrypt-file.1.adoc
@@ -1,0 +1,44 @@
+CLEVIS-ENCRYPT-FILE(1)
+======================
+:doctype: manpage
+
+
+== NAME
+
+clevis-encrypt-file - Encrypts using a file policy
+
+== SYNOPSIS
+
+*clevis encrypt file* CONFIG < PT > JWE
+
+== OVERVIEW
+
+The *clevis encrypt file* command encrypts using a file policy.
+Its only argument is the JSON configuration object.
+
+Encrypting data using the file pin works like this:
+
+    $ clevis encrypt file '{"name":"/path/to/file"}' < PT > JWE
+
+The given file must not exist yet.
+
+To decrypt the data, just pass it to the *clevis decrypt* command:
+
+    $ clevis decrypt < JWE > PT
+
+== CONFIG
+
+This command uses the following configuration properties:
+
+* *name* (string) :
+  The name to the file where the jwk is stored (REQUIRED)
+
+== BUGS
+
+Requires that directories for that file already exist.
+
+Rather for educational purposes.
+
+== SEE ALSO
+
+link:clevis-decrypt.1.adoc[*clevis-decrypt*(1)]
--- a/src/pins/file/clevis-decrypt-file
+++ b/src/pins/file/clevis-decrypt-file
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+set -eu
+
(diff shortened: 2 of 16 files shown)


src/pins/file/clevis-encrypt-file.1.adoc
```diff
@@ -0,0 +1,44 @@
...
+== CONFIG
+
+This command uses the following configuration properties:
+
+* *name* (string) :
+  The name to the file where the jwk is stored (REQUIRED)
+
+== BUGS
+
+Requires the directories for that file already exists.
```
sarroutbi, line 38: I would rewrite this as:
"Requires that directories for that file already exist"

sarroutbi (commented):
Minor change in documentation suggestion. Rest of changes LGTM.

cbiedl:
Certainly. Not being a native English speaker, such glitches do happen.
Do you want an updated pull request? Might be less work if you just do it on your own.

cbiedl:
Okay, found some minor issues so I've updated my branch (also, rebased it on top of HEAD). This resulted in some warnings/errors from CI I fail to understand. Feel free to take it as-is.

sarroutbi:
> Okay, found some minor issues so I've updated my branch (also, rebased it on top of HEAD). This resulted in some warnings/errors from CI I fail to understand. Feel free to take it as-is.

Thanks for the PR. The differential shellcheck issue looks like a false positive, so no need to worry for that.

zi4k4j74yn:
Is there any way one can assist to getting this merged?

sarroutbi:
Changes LGTM. @sergio-correia: can you please provide feedback when possible?

sarroutbi (approved):
Change LGTM. @sergio-correia : can you please provide your feedback when possible?

sarroutbi:
@sergio-correia : can you please provide your feedback when possible? I would prefer a second opinion. For me, this LGTM

sergio-correia (approved):
This looks good to me as well. Thanks, @cbiedl.

oldium:
FYI this will not be easy to package for distributors.

For Fedora-based distros this is straight-forward, because it uses just Dracut, so installing `clevis-decrypt-file` in `module-setup.sh` unconditionally is fine.

For Debian-based distros this will not be possible, because you can have either initramfs-tools or Dracut. The support for `initramfs-tools` is covered by `clevis-initramfs` package, support for Dracut is handled by clevis-dracut package.

To make this working in Debian, the hook scripts need to check if the clevis pin files exist prior to installing the full support.

For Dracut `module-setup.sh`, something like this should do the trick (both for `template` and for `file` pins):

`module-setup.sh.in`:
```
check() {
    require_binaries clevis-decrypt-@pin@ || return 1
}
```

For initramfs-tools, something like this should make it working:

```
if [ -x @bindir@/clevis-decrypt-@pin@ ]; then
    copy_exec @bindir@/clevis-decrypt-@pin@ || die 1 "@bindir@/clevis-decrypt-@pin@ not found"
fi
```

PR #204, opened 2020-05-24
Implement a pin to store the key on a block device. Closes: #185

cbiedl:
The "medium" pin allows storing the jwk in a file on a given block
device.


src/pins/medium/clevis-decrypt-medium
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
```
sarroutbi, line 1: To keep homogeneous scripts with similar shebang, this should be kept as
```bash
#!/bin/sh
```

PR #205, opened 2020-05-26
[wip] luks: get rid of bashisms in clevis-luks-common-functions

sergio-correia:
As suggested in #202, let's try to define some sort of style guide
to try to improve portability and consistency across the scripts.

The suggestions include:

- getting rid of bashisms, at least for scripts running in early
  userland
- running with "set -e" and "set -u"
- using $() instead of backticks
- using long form of `jose' commands, for better readability

This commit converts clevis-luks-common-function and its user
clevis-luks-list.

As of now, shellcheck complains of the following:
"SC2039: In POSIX sh, 'local' is undefined."

This can be muted with `shellcheck -eSC2039'

### DIFF
--- a/src/luks/clevis-luks-common-functions
+++ b/src/luks/clevis-luks-common-functions
@@ -1,4 +1,5 @@
-#!/bin/bash -e
+#!/bin/sh
+set -eu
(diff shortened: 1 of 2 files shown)


sergio-correia:
Closing this for now.

PR #210, opened 2020-07-01
initramfs - fix iface name

tavril:
This is a proposal for https://github.com/latchset/clevis/issues/209


sergio-correia:
Thanks, merged.

PR #220, opened 2020-08-12
pins: improve error reporting for sss pin

sergio-correia:
clevis-encrypt-sss would return 0 (and a JWE, without the pins that failed)
even if the encryption operation failed, when using the sss pin, such as:

echo test | clevis encrypt sss '{"t":1, "pins":{"tang":[{"url":"foo bar"}]}}'
Unable to fetch advertisement: 'foo bar/adv/'!
eyJhbGciOiJkaXIiLCJjbGV2aXMiOnsicGluIjoic3NzIiwic3NzIjp7Imp3ZSI6WyIiXSwicCI6Inlqald0bkM2ZkwtNDhIbTd5Yy1fVmlwWWxTTHZ1ZlliZnRaejFjLW5ucU0iLCJ0IjoxfX0sImVuYyI6IkEyNTZHQ00ifQ..rM6IMXjNIN3uybW1.J5jAtiU.61x_fjXJC1uCjljTnkkdCw

echo $?
0

This is particularly problematic when using clevis luks bind, as it would
appear the operation succeeded, while in practice it did not, which
means we would be unable to unlock this device. E.g.:

clevis luks bind -d device sss '{"t":1, "pins":{"tang":[{"url":"foo bar"}]}}' <<< password
Unable to fetch advertisement: 'foo bar/adv/'!

echo $?
0

clevis luks list -d device
1: sss '{"t":1,"pins":{}}'

clevis luks unlock -d device
Nothing to read on input.

We now check the exit code of the pins encryption, when doing an sss
encrypt, and indicate errors correctly, so that we avoid this situation.

After the fix:

echo test | clevis encrypt sss '{"t":1, "pins":{"tang":[{"url":"foo bar"}]}}'
Unable to fetch advertisement: 'foo bar/adv/'!

echo $?
1

Also:
clevis luks bind -d device sss '{"t":1, "pins":{"tang":[{"url":"foo bar"}]}}' <<< password
Unable to fetch advertisement: 'foo bar/adv/'!

echo $?
1

clevis luks list -d device
[no output]

### DIFF
--- a/src/luks/tests/bad-sss
+++ b/src/luks/tests/bad-sss
@@ -1,0 +1,40 @@
+#!/bin/bash -ex
+# vim: set ts=8 shiftwidth=4 softtabstop=4 expandtab smarttab colorcolumn=80:
+#
+# Copyright (c) 2020 Red Hat, Inc.
+# Author: Sergio Correia <scorreia@redhat.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+TEST=$(basename "${0}")
+. tests-common-functions
+
+on_exit() {
+    [ -d "${TMP}" ] && rm -rf "${TMP}"
+}
+
+trap 'on_exit' EXIT
+
+TMP="$(mktemp -d)"
+
+CFG='{"t":1, "pins":{"tang":[{"url":"foo bar"}]}}'
+
+# LUKS1.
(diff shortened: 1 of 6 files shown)


src/pins/sss/meson.build
```diff
@@ -18,8 +18,9 @@ if jansson.found() and libcrypto.found()
   mans += join_paths(src, 'clevis-encrypt-sss.1')
 
   env = environment()
-  env.append('PATH',
+  env.prepend('PATH',
```
jlebon, line 21: Is this change related to the rest of the patch? It looks like it's so that we use the in-tree scripts instead of the already installed ones? Might be cleaner as a separate commit with its own commit message.

src/luks/tests/bad-sss
```diff
@@ -0,0 +1,42 @@
...
+
+TEST=$(basename "${0}")
+. tests-common-functions
+
+on_exit() {
+    [ -d "${TMP}" ] && rm -rf "${TMP}"
+}
+
+trap 'on_exit' EXIT
+trap 'exit' ERR
```
jlebon, line 29: Is this line actually needed? Errors should cause exits already in `set -e` mode.

sergio-correia: Good catch. It's not required, indeed. Let me remove this.

src/pins/sss/meson.build
sergio-correia: You are right. Let me split this part.

PR #221, opened 2020-08-13
Small fixes

sergio-correia:
luks: improve error message when bind is given an invalid PIN
pins: update tang test so that it runs in all platforms
travis: install systemd in debian/ubuntu containers

### DIFF
--- a/.travis/install
+++ b/.travis/install
@@ -27,7 +27,7 @@ debian:*|ubuntu:*)
     while ! apt-get -y install ${COMMON} \
         build-essential pkg-config libssl-dev libjansson-dev libjose-dev \
         luksmeta libluksmeta-dev libpwquality-tools libglib2.0-dev \
-        libudisks2-dev libaudit-dev; do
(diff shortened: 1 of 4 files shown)


src/luks/clevis-luks-common-functions
```diff
@@ -309,6 +309,14 @@ clevis_luks_unlock_device() {
...
     done
+    return 1
+}
 
+# findexe() finds an executable.
+findexe() {
+    while read -r -d: path; do
+        [ -f "${path}/${1}" ] && [ -x "${path}/${1}" ] && \
+          echo "${path}/${1}" && return 0
+    done <<< "${PATH}:"
```
jlebon, line 320: Hmm, why not `command -v ${name}`?

sergio-correia: Good point. We have this function somewhere else in the code and I copied it here. `command -v` would be much more readable here.

PR #223, opened 2020-08-25
travis: attempt to update keys on rawhide

sergio-correia:
Ugly workaround in an attempt to make things work in rawhide after
branching.


sergio-correia:
Closing this is it does seem to be required for now.

PR #225, opened 2020-08-27
Jump forward to rewritten TPM2 pin if available

puiterwijk:
Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

### DIFF
--- a/src/pins/tpm2/clevis-decrypt-tpm2
+++ b/src/pins/tpm2/clevis-decrypt-tpm2
@@ -18,6 +18,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+# First try to use the new version of the PIN implementation
+if command -v clevis-pin-tpm2 >/dev/null;
+then
+    exec clevis-pin-tpm2 decrypt $@
+fi
+
 # The owner hierarchy is the one that should be used by the Operating System.
 auth="o"
 
--- a/src/pins/tpm2/clevis-encrypt-tpm2
+++ b/src/pins/tpm2/clevis-encrypt-tpm2
@@ -18,6 +18,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+# First try to use the new version of the PIN implementation
+if command -v clevis-pin-tpm2 >/dev/null;
+then
+    exec clevis-pin-tpm2 encrypt "$@"
+fi
+
 SUMMARY="Encrypts using a TPM2.0 chip binding policy"
 # The owner hierarchy is the one that should be used by the Operating System.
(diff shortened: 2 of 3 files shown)


martinezjavier (approved):
Looks good to me! Thanks a lot for this.

src/pins/tpm2/clevis-decrypt-tpm2
```diff
@@ -18,6 +18,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+# First try to use the new version of the PIN implementation
+if $(which clevis-pin-tpm2 >/dev/null 2>&1);
```
sergio-correia, line 22: Please, use `command` to check if `clevis-pin-tpm2` is available, so that we do not have to ship `which` in the initramfs. Probably `if command -v clevis-pin-tpm2 >/dev/null; then` should do.

puiterwijk: Fixed, sorry, I thought `which` was also a bash built-in.

src/pins/tpm2/clevis-decrypt-tpm2
```diff
@@ -18,6 +18,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+# First try to use the new version of the PIN implementation
+if command -v clevis-pin-tpm2;
```
sergio-correia, line 22: You need to redirect stdout to `/dev/null`, or else it will print the path of `clevis-pin-tpm2`, when it is found :)

sergio-correia:
Hm. now test `pin-tpm2` is failing, when trying to parse this configuration: `'{"pcr_ids": "4,  16"}'`

```
++ clevis encrypt tpm2 '{"pcr_ids": "4,  16"}'
Error during parsing operation: Error: Unparseable string int
```

puiterwijk:
@sergio-correia yeah, I just (an hour ago) pushed a commit to the upstream of the pin to fix handling those formats: https://github.com/fedora-iot/clevis-pin-tpm2/commit/e2f32d0fa0ea5917049e7c852945dbfcdb4e6be1
This commit is in v0.1.3 of the pin, which should make its way to Fedora soon.
In order to run the test for now, you can git clone the repository, `cargo build` and set the `$pwd/target/debug` in your `$PATH`.

sergio-correia:
@puiterwijk: ah, thanks for the clarification. Testing with https://github.com/fedora-iot/clevis-pin-tpm2/commit/e2f32d0fa0ea5917049e7c852945dbfcdb4e6be1 works, so this looks good. Thanks again for this work.

PR #239, opened 2020-09-23
clevis-luks-askpass.service: use Type=simple

jlebon:
Let's use `Type=simple` since `clevis-luks-askpass -l` is long-running
and not "one-shot".

Noticed this while debugging something else. It doesn't make any
practical difference but could in the future. With `oneshot`, the
service will be stuck in `activating` which means that systemd won't be
able to start any follow-up services which explicitly do
`After=clevis-luks-askpass.service`.


sergio-correia (approved):
Nice catch, thanks.

PR #257, opened 2020-11-11
pins/tpm2: add support for tpm2-tools 5.X

diabonas:
A release candidate for tpm2-tools 5.0 [has been tagged](https://github.com/tpm2-software/tpm2-tools/releases/tag/5.0-rc0). Since we explicitly check the major version of the tools (to support the incompatible versions 3.X and 4.X), the allowed version range needs to be expanded.

tpm2-tools 5.0 consolidates all tools into a single busybox-style binary, so the preferred way to invoke the commands would be e.g. `tpm2 createprimary` instead of `tpm2_createprimary`. However, compatibility symlinks `tpm2_*` → `tpm2` are installed by default, so we keep the old syntax for tpm2-tools 5.0 to avoid creating another special case, since the option syntax has not changed (it should be stable since version 4).

tpm2-tools 3.X is deprecated, but unfortunately still packaged by a few Linux distributions, so keep supporting it for now at least.


sergio-correia (approved):
Thanks, it looks good.

PR #264, opened 2020-12-01
tang: default JWK thumbprint is now SHA-256 / deprecate SHA-1

sergio-correia:
During the encryption using tang, the clevis metadata stores the
thumbprint of the exchange key in the `kid' attribute. This thumbprint
uses jose default thumbprint hash algorithm -- SHA1 as of now.

When decrypting, we validate that the key with thumbprint represented by
`kid' is present in the advertisement. At this point, jose then uses its
default hash algorithm to do the checking. If jose default thumbprint
hash algorithm changes and we have a different default than what was used
to calculate `kid' at encryption time, this verification will fail, and
so does the decryption itself.

To handle a possible change in jose default thumbprint hash algorithm,
this verification was extended to try out every jose supported hash
algorithms, if the validation of `kid' fails with the default one.

### DIFF
--- a/src/pins/tang/clevis-decrypt-tang
+++ b/src/pins/tang/clevis-decrypt-tang
@@ -50,10 +50,36 @@ if ! kid="$(jose fmt -j- -Og kid -Su- <<< "$jhd")"; then
     exit 1
 fi
 
-if ! srv="$(jose fmt -j- -Og clevis -g tang -g adv -Oo- <<< "$jhd" \
-        | jose jwk thp -i- -f "$kid")"; then
+# Tang advertisement validation.
+if ! keys="$(jose fmt -j- -Og clevis -g tang -g adv -Oo- <<< "${jhd}")"; then
     echo "JWE missing required 'clevis.tang.adv' header parameter!" >&2
     exit 1
+fi
+
+# Check if the thumbprint we have in `kid' is in the advertised keys.
+CLEVIS_DEFAULT_THP_ALG=S256       # SHA-256.
+CLEVIS_DEFAULT_THP_LEN=43         # Length of SHA-256 thumbprint.
+CLEVIS_ALTERNATIVE_THP_ALGS=S1    # SHA-1.
+
+# Issue a warning if we are using a hash that has a shorter length than the
+# default one.
+if [ "${#kid}" -lt "${CLEVIS_DEFAULT_THP_LEN}" ]; then
+    echo "WARNING: tang using a deprecated hash for the JWK thumbprints" >&2
+fi
+
+if ! srv="$(jose jwk thp -i- -f "${kid}" -a "${CLEVIS_DEFAULT_THP_ALG}" \
+            <<< "${keys}")"; then
+    # `kid' thumprint not in the advertised keys, but it's possible it was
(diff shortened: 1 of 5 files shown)


sergio-correia:
This is related to https://github.com/latchset/jose/issues/86

src/pins/tang/clevis-decrypt-tang
```diff
@@ -50,12 +50,27 @@ if ! kid="$(jose fmt -j- -Og kid -Su- <<< "$jhd")"; then
...
     echo "JWE missing required 'clevis.tang.adv' header parameter!" >&2
     exit 1
 fi
 
+# Check if the thumbprint we have in `kid' is in the advertised keys.
+if ! srv="$(jose jwk thp -i- -f "${kid}" <<< "${keys}")"; then
+    # `kid' thumprint not in the advertised keys, but it's possible jose used
+    # a different default algorithm for calculating its thumbprint, so let's
+    # try out the supported algorithms to make sure `kid' really is not part
+    # of the advertised keys.
```
Jakuje, line 64: Would it make sense to derive the algorithm from the hash length and limit to sensible hashes as SHA1, SHA256 for now? This could limit the amount of attempts here (if jose supports more hash algorithms) and speed up things.

sergio-correia: Yeah, something like this would work, but why would you want to limit it to SHA1 and SHA256?

Jakuje: I am not sure what all hash types you are planning to support nor what hash types does jwk support at all, but as you are not transferring the hash type alongside of the hash and just trying all of them, it makes it prone to collisions especially if an attacker could modify the thumbprints and jwk supports md5 or other ancient stuff.
Limiting to known list of good algorithm can give you assurance what is actually being used. Using SHA1 is the current behavior, which is good enough for short future, SHA256 is something that is still possible by users to compare manually and good for quite long future, but anything bigger is hard to compare by users. I see this is automated comparison so it should not be an issue, but other uses might require manual comparison.

sergio-correia: These are good points, thanks for elaborating. I agree it is a good idea to limit it to SHA1 and SHA256 for now.

sergio-correia: Would you mind taking another look at the updated version?

Jakuje (approved):

ZelenyMartin (approved):
Does make sense to have some same parts of clevis-encrypt-tang and clevis-decrypt-tang detached to some common functions file like in clevis-luks scripts? Only think I see the same are the constants and cycle through the CLEVIS_ALTERNATIVE_THP_ALGS - this is definitely unnecessary. It's just minor comment.

sergio-correia:
> Does make sense to have some same parts of clevis-encrypt-tang and clevis-decrypt-tang detached to some common functions file like in clevis-luks scripts? Only think I see the same are the constants and cycle through the CLEVIS_ALTERNATIVE_THP_ALGS - this is definitely unnecessary. It's just minor comment.

It probably would make sense to do that, at some point. As you pointed out, the common parts are currently minimal, so I think it would be fine to keep it as is for now.

PR #275, opened 2020-12-25
Work around /dev/fd/X removal in systemd

cbiedl:
The <(command) notation in bash relies on the existence of /dev/fd/X,
these device nodes however are no longer created by udev/systemd and
therefore are not available in early userland, breaking unlocking.

Work around by re-writing the code using pipe.

systemd commit:
    https://github.com/systemd/systemd/commit/6b2229c6c60d0486

Debian bug report:
    https://bugs.debian.org/968518

Closes: #262

### DIFF
--- a/src/clevis-decrypt
+++ b/src/clevis-decrypt
@@ -46,7 +46,8 @@ if ! [ -t 0 ]; then
         exit 1
     fi
 
-    exec "$cmd" < <(echo -n "$hdr."; /bin/cat)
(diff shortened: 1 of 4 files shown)


dpantel:
`/bin/clevis-luks-common-functions` and possibly other scripts also need patching.  See #274.

sergio-correia (approved):

sergio-correia:
Thanks, merged.

PR #278, opened 2020-12-29
Add spaces ot [ invocation in "assume-yes" test

cbiedl:
Else the test will trigger messages like

    src/luks/tests/assume-yes: line 71: [foo: command not found


sergio-correia (approved):
Thanks for catching these.

PR #280, opened 2021-01-10
clevis-luks-bind: Fix error with password of spaces

ly4096x:
This fixes the problem with passwords that contain spaces at beginning or ending. Such a password will make the `existing_key` variable has an incorrect key and result into a "No key available with this passphrase.` or "Nothing to read on input." error.

### DIFF
--- a/src/luks/clevis-luks-bind
+++ b/src/luks/clevis-luks-bind
@@ -98,8 +98,8 @@ fi
 existing_key=
 keyfile=
 case "${KEY}" in
-"") read -r -s -p "Enter existing LUKS password: " existing_key; echo >&2;;
- -) read -r -s -p "" existing_key
+"") IFS= read -r -s -p "Enter existing LUKS password: " existing_key; echo >&2;;
+ -) IFS= read -r -s -p "" existing_key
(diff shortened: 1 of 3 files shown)


src/luks/clevis-luks-bind
```diff
@@ -98,8 +98,8 @@ fi
 existing_key=
 keyfile=
 case "${KEY}" in
-"") read -r -s -p "Enter existing LUKS password: " existing_key; echo >&2;;
- -) read -r -s -p "" existing_key
+"") echo -n "Enter existing LUKS password: "; existing_key=$(head -1); echo >&2;;
```
sergio-correia, line 101: Something bad with this fix is that the password will be displayed as it is being typed. Perhaps a better way to fix this would be to clear IFS? Something like the following:

`IFS= read -r -s -p "Enter existing LUKS password: " existing_key; echo >&2;;`

ly4096x: Yes, you're right. This is a better solution. I have updated my code. Thank you!

sergio-correia (approved):
Looks good, thanks!

PR #282, opened 2021-01-13
ci: correct failing logic

ly4096x:
This fixes the auto test always succeeding due to the `cat` making fail return code being ignored.


sergio-correia (approved):

sergio-correia:
Thanks

PR #295, opened 2021-02-23
luks: replace `seq` with bash's builtin sequence expression

jlebon:
This gives the same result without the added dep.

Hit this in FCOS where we don't currently ship `seq` in the initramfs
(normally, the clevis dracut module should've installed it if it
requires it). We could do that of course, but it's even better if we can
avoid it entirely.


sergio-correia (approved):
Thanks for catching this.

BarryHuffman:
Does this remove agnostic shell use? (dash/zsh/fish)

PR #297, opened 2021-03-15
Fix for -t option in clevis luks bind

jbaublitz:
@sergio-correia Here is the fix. I've tested it locally and added a test as we discussed.

### DIFF
--- a/src/luks/tests/bind-luks2
+++ b/src/luks/tests/bind-luks2
@@ -49,5 +49,5 @@ if ! clevis luks bind -d "${DEV}" -t "$TOKEN_ID" tang "${CFG}" <<< "${DEFAULT_PA
 fi
 
 if ! cryptsetup token export --token-id=5; then
-    error "${TEST}: Clevis did not add the LUKS2 token to the correct slot"
+    error "${TEST}: Clevis did not add the LUKS2 token to the correct slot." >&2
 fi
--- a/src/luks/tests/bind-luks2
+++ b/src/luks/tests/bind-luks2
@@ -48,6 +48,6 @@ if ! clevis luks bind -d "${DEV}" -t "$TOKEN_ID" tang "${CFG}" <<< "${DEFAULT_PA
     error "${TEST}: Binding is expected to succeed when given a correct (${DEFAULT_PASS}) password." >&2
 fi
 
-if ! cryptsetup token export --token-id=5; then
+if ! cryptsetup token export --token-id=5 "${DEV}"; then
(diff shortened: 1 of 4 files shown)


jbaublitz:
I'm having a little trouble with the testing infrastructure but hopefully this helps!

src/luks/tests/bind-luks2
```diff
@@ -47,3 +47,7 @@ new_device "luks2" "${DEV}"
 if ! clevis luks bind -d "${DEV}" -t "$TOKEN_ID" tang "${CFG}" <<< "${DEFAULT_PASS}"; then
     error "${TEST}: Binding is expected to succeed when given a correct (${DEFAULT_PASS}) password." >&2
 fi
+
+if ! cryptsetup token export --token-id=5; then
```
sergio-correia, line 51: The issue is here: https://github.com/latchset/clevis/pull/297/checks?check_run_id=2116394266#step:7:5079
You are missing the device in the `cryptsetup` command.

src/luks/tests/meson.build
```diff
@@ -1,5 +1,7 @@
 # We use jq for comparing the pin config in the clevis luks list tests.
 jq = find_program('jq', required: false)
+# We use cryptsetup for testing LUKS2 binding and saving the token in a given token slot.
+cryptsetup = find_program('cryptsetup', '/usr/sbin/cryptsetup', required: true)
```
sergio-correia, line 4: This should not be required.

src/luks/tests/bind-luks2
jbaublitz: Whoops! Sorry about that. This should be resolved.

jbaublitz:
The most recent test failures were what I first saw which is why I made the edits. They don't appear to be related to my changes as there are errors like `cryptsetup not found`.

sergio-correia:
> The most recent test failures were what I first saw which is why I made the edits. They don't appear to be related to my changes as there are errors like `cryptsetup not found`.

Yeah, I see it now. It looks like `cryptsetup` is not installed in rawhide anymore -- we were lucky before that it was installed as a dependency of something else --, so let's be explicit and add it there: https://github.com/latchset/clevis/blob/master/.github/workflows/install-dependencies#L36
I will update the spec file later to add it as a build dependency.

sergio-correia:
@jbaublitz: please, rebase, as this build issue should have been fixed by #298

jbaublitz:
@sergio-correia Okay, rebased!

sergio-correia (approved):

sergio-correia:
Thanks!

PR #328, opened 2021-07-19
tpm2: improve validation of PCRs in clevis-encrypt-tpm2

sergio-correia:
After parsing the PCR IDs, validate the combination of pcr_bank and
the PCRs, so that we can better handle invalid situations.

Also update our tpm2 tests to skip some tests if the configuration is
not valid.

Fixes: #327

### DIFF
--- a/src/pins/tpm2/pin-tpm2
+++ b/src/pins/tpm2/pin-tpm2
@@ -46,6 +46,35 @@ tpm2_available() {
         echo "The ${TPM2TOOLS_DEVICE_FILE} device must be readable and writable!" >&2
         return 1
     fi
+
+    local _tpm2tools_info="$(tpm2_createprimary -v)"
+    local _match='version="(.)\.'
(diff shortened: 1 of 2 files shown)


src/pins/tpm2/pin-tpm2
```diff
@@ -46,6 +46,32 @@ tpm2_available() {
         echo "The ${TPM2TOOLS_DEVICE_FILE} device must be readable and writable!" >&2
         return 1
     fi
+
+    local TPM2TOOLS_INFO="$(tpm2_createprimary -v)"
```
sarroutbi, line 50: as this is a local variable, it might be defined as _tpm2tools_info for homogeneity with rest of the code

src/pins/tpm2/pin-tpm2
```diff
@@ -46,6 +46,32 @@ tpm2_available() {
...
+    local match='version="(.)\.'
```
sarroutbi, line 51: as this is a local variable, it might be defined as _match for homogeneity with rest of the code

PR #337, opened 2021-10-19
luks: improve error handling when performing encryption

sergio-correia:


sergio-correia:
> Just a comment, why this: if ....; then ... fi Instead of this?: if ....; then .... fi
> 
> I consider the second is more readable, but I guess it is very subjective

I would be fine with the second option, but the rest of the code uses the first one, so I am keeping it consistent.

PR #348, opened 2021-11-05
Verify that tpm2_flushcontext is installed and added to initramfs

jpds:
This is required as c6fc63fc055c18927decc7bcaa07821d5ae37614 added the binary to the decrypt scripts.


sergio-correia (approved):
Thanks.

PR #353, opened 2021-12-01
feat: rename the test pin to null pin

runcom:
Fixes https://github.com/latchset/clevis/issues/352

Rename the test pin to null pin (didn't move it to the top level pins directory as it's still tied to sss but I guess it makes no difference, does it?) - sorry if something is wrong, first time contributing here

Signed-off-by: Antonio Murdaca <runcom@linux.com>


sergio-correia:
This looks OK, but you want to ship the `null` pin, no? If so, you should add something like the following in the `sss` `meson.build` file:

```
  bins += join_paths(meson.current_source_dir(), 'clevis-encrypt-null')
  bins += join_paths(meson.current_source_dir(), 'clevis-decrypt-null')
```

Also, don't you need a dracut module as well?

runcom:
> This looks OK, but you want to ship the `null` pin, no? If so, you should add something like the following in the `sss` `meson.build` file:
> 
> ```
>   bins += join_paths(meson.current_source_dir(), 'clevis-encrypt-null')
>   bins += join_paths(meson.current_source_dir(), 'clevis-decrypt-null')
> ```
> 
> Also, don't you need a dracut module as well?

thanks Sergio, I've pushed a change for that - can you take a look?
also, yeah, we need also the dracut module - I thought that's on the RPM spec side of the world,isn't it?

src/pins/sss/meson.build
```diff
@@ -28,8 +28,12 @@ if jansson.found() and libcrypto.found()
     separator: ':'
   )
 
+  bins = []
```
sergio-correia, line 31: Please, remove this line, as this variable is defined elsewhere, and with this you are overriding its value.

sergio-correia:
> also, yeah, we need also the dracut module - I thought that's on the RPM spec side of the world,isn't it?

We need something similar to what we have for the other pins, i.e., a directory under `src/luks/systemd/dracut` with a module that installs the decrypt program.

src/pins/sss/meson.build
runcom: fixed thanks!

runcom:
> We need something similar to what we have for the other pins, i.e., a directory under `src/luks/systemd/dracut` with a module that installs the decrypt program.

done!

sergio-correia:
Thanks, looks good. A couple of last changes:
1) please add a `subdir('clevis-pin-null')` in `src/luks/systemd/dracut/meson.build` so that it will include the new directory you just created
2) could you please also add `copy_exec` call for the decrypt program similar to the other ones we have already in `src/initramfs-tools/hooks/clevis.in`, so that it's installed also for `initramfs-tools` users?

runcom:
> Thanks, looks good. A couple of last changes:

all done now!

sergio-correia (approved):

sergio-correia:
Thanks, merged!

runcom:
@sergio-correia can we get a build in brew for both 8.6 and 9?

runcom:
> Thanks, merged!

hey @sergio-correia should we build clevis again for fedora too?

sergio-correia:
> > Thanks, merged!
> 
> hey @sergio-correia should we build clevis again for fedora too?

Sorry, I missed this. Yes, I will do that.

PR #357, opened 2022-01-04
luks: fix typo when adding a pending device

tmzullinger:
In `clevis_devices_to_unlock()` we use `${clevis_devices}` to track the
devices which need to be processed.  When we are unable to get the
device, we add it to the list and try to activate on a subsequent trip
through the loop.

Use `${crypt_device}` rather than `${dev}` as the latter may well be
empty.  This crept in with the refactoring in 7c17448 (systemd: rework
clevis-luks-askpass for improved reliability, 2020-08-26).

Sergio confirmed that `${crypt_device}` was the intended variable in
response to the question in 0589c14 (luks: ignore empty & comment lines
in crypttab, 2021-11-03)¹.

¹ https://github.com/latchset/clevis/pull/355#pullrequestreview-843477229


sergio-correia (approved):
Thanks!

PR #359, opened 2022-01-12
src/luks/systemd/dracut/meson.build: Empty line at EOF

travier:
Minor fix for https://github.com/latchset/clevis/pull/353


sergio-correia (approved):
Nice catch, thanks!

PR #374, opened 2022-06-05
initramfs: add support for mkinitcpio initramfs generator

### LINKED ISSUE #334: [Feature request]  Add mkinitcpio hook support.
I use Arch Linux and a someone I've talked with on their IRC channel stated that the clang package for Arch Linux doesn't have mkinitcpio support because upstream didn't. For Arch Linux mkinitcpio is the default way of generating a init ram disk including adding any modules and features you may want to have at boot time.  https://wiki.archlinux.org/title/Mkinitcpio

This is a mkinitcpio install and hook I found from the Arch Linux forums that would enable clang at boot but I have no idea if this is totally correct. https://bbs.archlinux.org/viewtopic.php?id=230948

clevis-mkinitcpio-install:
```
#!/bin/bash

build() {
    add_binary "clevis-decrypt-http"
    add_binary "clevis-decrypt-tang"
    add_binary "clevis-decrypt-sss"
    add_binary "clevis-decrypt"
    add_binary "luksmeta"
    add_binary "clevis"
    add_binary "jose"
    add_binary "curl"
    add_binary "bash"
    add_runscript
}

help() {
    cat <<HELPEOF
This hook will attempt to unlock LUKS volumes using data stored in the header
by clevis and luksmeta. Use this hook in combination with any early userspace
networking hook, such as mkinitcpio-netconf or mkinitcpio-ppp. It also requires
mkinitcpio-utils for the encryptssh hook.

An example usage would be to have 'netconf clevis encryptssh' added before your
filesystems hook. You also need to configure clevis unlocking by using the
'clevis bind luks' command on your luks partition.
HELPEOF
}
```

clevis-mkinitcpio-hook: 
```
#!/usr/bin/bash

run_hook ()
{
    clevis_loop &
    echo $! > /.clevis.pid
}

clevis_loop()
{
    while ! [ -c /dev/mapper/control -a -f /.cryptdev -a -f /.cryptname -a -f /.cryptargs ];
do
        sleep 0.1
    done

    CRYPTDEV=$(cat /.cryptdev)
    CRYPTNAME=$(cat /.cryptname)
    UUID=cb6e8904-81ff-40da-a84a-07ab9ab5715e
    luksmeta show -d "$CRYPTDEV" | while read -r slot state uuid; do
        [ "$state" != "active" ] && continue
        [ "$uuid" != "$UUID" ] && continue

        if ( luksmeta load -d "$CRYPTDEV" -s $slot -u $UUID | clevis decrypt \
            | cryptsetup luksOpen $(cat /.cryptdev) $CRYPTNAME $(cat /.cryptargs) ); then
            echo > /.done
            echo -e "\n\n$CRYPTNAME sucessfully decrypted via clevis."
            killall cryptsetup
            break
        fi
    done
    rm /.clevis.pid
}

run_cleanuphook ()
{
    if [ -f /.clevis.pid ]; then
        kill $(/.clevis.pid)
        rm /.clevis.pid
    fi
}
```

diabonas:
[mkinitcpio](https://wiki.archlinux.org/title/Mkinitcpio) is the default mkinitcpio used by Arch Linux. It can be extended through custom hooks, so provide one that installs all the necessary files for `clevis luks unlock` to function. The device to be unlocked is configured through the [`cryptdevice`](https://wiki.archlinux.org/title/Dm-crypt/System_configuration#cryptdevice) kernel command line option in the same way the built-in `encrypt` mkinitcpio hook works.

Resolves: #334

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,34 @@ sudo update-initramfs -u -k 'all'
 
 Upon reboot it will behave exactly as if using Dracut.
 
+### Unlocker: mkinitcpio
+
+When using Clevis with mkinitcpio, you will need to add the `clevis` hook
+before the `encrypt` hook in `/etc/mkinitcpio.conf`:
+
+```bash
+HOOKS=( ... clevis encrypt ... )
+```
+
+Then run
+
+```bash
+$ sudo mkinitcpio -P
+```
+
+in order to regenerate the initramfs. The device to be unlocked is configured
+on the kernel command line using the `cryptdevice` option in the same way as
+for the default `encrypt` hook:
+
+```bash
+cryptdevice=/dev/sda1:root
+```
+
+In order to use a Tang pin you will need to enable networking in the initramfs
+by adding and configuring the [`net`
+hook](https://wiki.archlinux.org/title/Mkinitcpio#Using_net) *before* the
+`clevis` hook.
+
 #### Unlocker: UDisks2
 
 Our UDisks2 unlocker runs in your desktop session. You should not need to
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,2 +1,3 @@
 option('user', type: 'string', value: 'clevis', description: 'Unprivileged user for secure clevis operations')
 option('group', type: 'string', value: 'clevis', description: 'Unprivileged group for secure clevis operations')
+option('mkinitcpiodir', type: 'string', description: 'mkinitcpio hooks directory')
--- a/src/meson.build
+++ b/src/meson.build
@@ -2,6 +2,7 @@ subdir('bash')
 subdir('luks')
 subdir('pins')
 subdir('initramfs-tools')
+subdir('mkinitcpio')
 
 bins += join_paths(meson.current_source_dir(), 'clevis-decrypt')
 mans += join_paths(meson.current_source_dir(), 'clevis-decrypt.1')
--- a/src/mkinitcpio/clevis.hooks
+++ b/src/mkinitcpio/clevis.hooks
@@ -1,0 +1,10 @@
+#!/usr/bin/ash
+
+run_hook() {
+    IFS=: read cryptdev cryptname cryptoptions <<EOF
+$cryptdevice
+EOF
+    if resolved=$(resolve_device "${cryptdev}" ${rootdelay}); then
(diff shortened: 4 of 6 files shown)


diabonas:
The CI failure appears to be unrelated, this pull request does not touch any actual code.

richie-tt:
Hi, is there any possibility of further consideration of this Pull Request?

@sergio-correia  @sarroutbi @oldium ?

sarroutbi:
Sorry, but I don't have enough knowledge about `mkinitcpio`. It LGTM, but I would prefer @sergio-correia , @oldium to share their opinion

oldium:
`mkinitcpio` is Arch-specific tool, something like Debian's `mkinitramfs`. According to Arch [documentation](https://wiki.archlinux.org/title/Dm-crypt/System_configuration), it supports old-fashioned scripted unlocking (via `encrypt` module), but also supports systemd startup, like Dracut with systemd (but I have not seen it running, so I just _guess_ it is similar), for more standardized `cryptsetup` loading.

The current mkinitramfs/Dracut way to unlock the devices is to hook into the password-providing paths and provide a correct password and let the reponsible code to pass the correct options to `cryptsetup` (like `--allow-discards` for SSDs).

This implementation is limited in the way that it specifically requires running before `encrypt` module (which does the old-fashioned unlocking without systemd), parses the `encrypt`'s `cryptdevice` kernel cmdline parameter, and unlocks the device by a call to `clevis unlock`. No options are passed to cryptsetup.

I like the way how we currently hook into the password workflow just to provide the password and let the responsible scripts to unlock the device and provide correct cryptsetup options. We are able to hook into the cryptsetup askpass, plymouth ask-for-password, systemd-ask-password, but also to a key-file reading paths (which I added in this [commit](https://github.com/latchset/clevis/pull/462/commits/77678cf5392ce9be2da1bbdaf27bfd4d5c181c9e) - not yet merged). We are rather flexible in this.

So it depends on how to proceed. I think the most future proof is to pass the unlocking to the system scripts specifically written for unlocking, and hook into the password-providing path only. For me the best of all worlds is systemd, because it allows adding TPM1 unlocking (it requires Trousers daemon running), all others require manually starting the daemon and terminating it. But also non-systemd paths are doable, it just takes more time...

PR #380, opened 2022-07-16
Fix typo in error messages

bits01:


sarroutbi (approved):
Changes LGTM. Thanks for the fix

PR #383, opened 2022-08-16
luks-edit: remove unnecessary 2>/dev/null

RPigott:
This redirection is unnecessary and hides many interesting errors when rebinding
with clevis


sergio-correia (approved):
Thanks.

PR #390, opened 2022-11-02
sss: sss-encrypt must read pipe data until EOF is reached

### LINKED ISSUE #389: SSS encrypt produces an empty slot
I am running clevis regen and some drives are left with invalid clevis tokens. I use SSS and some jwe pins are empty strings, like this (pay attention to the last element of `jwe` array).

```
{"jwe":["token1data", "token2data",""],"p":"yita4nK0gX4oLGo3Crje19JgeGeb7jjOBb505qSUMws","t":3}
```

I tried to triage it and I guess this issue is related to the clevis sss.c code. In particular this one https://github.com/latchset/clevis/blob/23f9bf7c90ae523570d6ef968aef918a5026b290/src/pins/sss/clevis-encrypt-sss.c#L118

```c
while (!feof(pipe)) {
    char buf[1024] = {};
    json_t *tmp = NULL;
    size_t rd = 0;

    rd = fread(buf, 1, sizeof(buf), pipe);
    if (ferror(pipe)) {
        fclose(pipe);
        return NULL;
    }

    tmp = json_pack("s+%", json_string_value(jwe), buf, rd);
    if (!tmp) {
        fclose(pipe);
        return NULL;
    }

    json_decref(jwe);
    jwe = tmp;
}
```

I do not fully understand how this code supposes to work. What if `fread()` is called before the pin program produces any output? In this case, json_pack() will pack an empty string and return it to the caller. It does not sound right to me.

I also looked at the decryption codepath here https://github.com/latchset/clevis/blob/23f9bf7c90ae523570d6ef968aef918a5026b290/src/pins/sss/clevis-decrypt-sss.c#L229 and it does pipe reading in a correct way - it basically reads/appends to a buffer until the pipe is closed. Thus the pipe is fully read before its output gets processed.

For me, it looks like encrypt should use the same codepath as decrypt.

anatol:
With current implementation sss-encrypt opens a pipe to the child pin, then tries to read this pipe *only once*. This might return only partial data or not data at all in case if the child pin response is slow.

Fix it by reading the pipe until it is closed, and only then process the data.

Also increase the response buffer from 1024 bytes to 4096 bytes to accound for larger subpin responses that can be larger than 1024 bytes.

Fixes #389

### DIFF
--- a/src/pins/sss/clevis-encrypt-sss.c
+++ b/src/pins/sss/clevis-encrypt-sss.c
@@ -115,26 +115,34 @@ encrypt_frag(json_t *sss, const char *pin, const json_t *cfg, int assume_yes)
     if (!pipe)
         return NULL;
 
+    char buf[4096] = {};
+    size_t rd = 0;
+    json_t *tmp = NULL;
     while (!feof(pipe)) {
-        char buf[1024] = {};
-        json_t *tmp = NULL;
-        size_t rd = 0;
+        char tmp_buf[4096] = {};
+        size_t tmp_rd = 0;
 
-        rd = fread(buf, 1, sizeof(buf), pipe);
+        tmp_rd = fread(tmp_buf, 1, sizeof(tmp_buf), pipe);
         if (ferror(pipe)) {
             fclose(pipe);
             return NULL;
         }
-
-        tmp = json_pack("s+%", json_string_value(jwe), buf, rd);
-        if (!tmp) {
+        if (rd + tmp_rd > sizeof(buf)) {
             fclose(pipe);
+            fprintf(stderr, "sss: read buffer overflow\n");
             return NULL;
         }
-
-        json_decref(jwe);
-        jwe = tmp;
+        memcpy(buf + rd, tmp_buf, tmp_rd);
+        rd += tmp_rd;
     }
+    tmp = json_pack("s+%", json_string_value(jwe), buf, rd);
+    if (!tmp) {
+        fclose(pipe);
+        return NULL;
+    }
+
+    json_decref(jwe);
+    jwe = tmp;
 
     fclose(pipe);
     waitpid(pid, &status, 0);


RPigott:
This is way more wrong than the current function. The only thing wrong with the current implementation is that it calls feof before reading from the stream — it should loop on fread and test feof only if the number of read bytes is less than requested.

This version eliminates the read-write loop and introduces a superfluous read loop around just fread, so it fails to accomodate any jwe larger than the temporary buffer. fread already will try to read all the requested bytes from the stream except in case of EOF or io error, so there is no need to loop there.

If you believe a signal is causing your error, you'd have better luck looking at the write end of this pipe, which does use raw `write`s _and_ neglects to check for EINTR before erring out:
https://github.com/latchset/clevis/blob/399919908860b354e793d66eb4a55f6f95bbac78/src/pins/sss/sss.c#L371-L375

anatol:
Hi Ronan,

Thank you for your feedback.

I see a lot of confusion in your comment, so let me clarify the PR, so hopefully, it becomes clear what we are trying to achieve here.

> This is way more wrong than the current function. The only thing wrong with the current implementation is that it calls feof before reading from the stream

Not really. The problem is that this loop body is effectively executed only once. It means that if pipe contains more incoming data (e.g. when the other side of the pipe is slow), then `json_pack` will try to parse incomplete/incorrect data. We found that it leads to clevis binding data corruption and essentially makes the binding unusable. It is a severe problem that needs to be fixed.

> This version eliminates the read-write loop

There is no `write()` involved in this patch, neither before my change, nor after.

> and introduces a superfluous read loop

There is no introduction of new loops. There was only 1 `while` loop before, and this loop stays with my patch.

> so it fails to accomodate any jwe larger than the temporary buffer

I am not sure what you really mean here. Are you trying to tell that 4096 bytes buffer is insufficient and needs to be bigger? Or it is something else. Please clarify.

> fread already will try to read all the requested bytes from the stream except in case of EOF or io error, so there is no need to loop there.

There are more cases when fread can return less bytes than requested. Please see the manpage for more information.

anatol:
>  look at the usage of json_pack("s+%", ...)

D'oh, I did not realize what the plus sign semantic is. Now the original code starts to make sense - `jwe` acts as an incomplete json data buffer. It is not a typical way to build up the data buffer, and it confused me.

In this case, my PR indeed makes no functional change and can be dropped.

> I have read both the manpage and the GNU Texinfo manual; these are indeed the only cases.

Here is a quote from `read` syscal (used as an underlying mechanism for fread) that explicitly mentions pipes:
```
On success, the number of bytes read is returned (zero indicates
end of file), and the file position is advanced by this number.
It is not an error if this number is smaller than the number of
bytes requested; this may happen for example because fewer bytes
are actually available right now (maybe because we were close to
end-of-file, or because we are reading from a pipe, or from a
terminal), or because read() was interrupted by a signal.
```

Or in other words, one cannot assume that the pipe blocks the read until EOF is received.

RPigott:
I'm aware the read syscall can be interrupted, but the pipe will not indicate EOF in those cases. `fread()` is not a syscall wrapper, it is a C standard "buffered io" function that reads from a _stream_. It's behavior is specified in §7.19.8.1 "The fread function" in the C standard, and that matches the description from `fread(3)`. It is mandated to read the number of bytes requested except in the cases above.

Obviously, on Linux, to achieve the behavior mandated by the standard it must loop, potentially calling read more than once. And it does. See the musl libc implementation here (b/c glibc is a mess): https://git.musl-libc.org/cgit/musl/tree/src/stdio/fread.c#n26.

What I was trying to point out before is that the parent process, when writing to the write end of this pipe, doesn't use the streams function `fwrite()`, but plain a `write()` call which is subject to the interruption behavior on Linux similar to `read()`. They loop to mitigate this, but unconditionally exit on error, even when the error may be benign (i.e. EINTR).

anatol:
Ack, thank you for the clarification @RPigott !

anatol:
Closing this PR.

It turns out the clevis data corruption is somewhere else. It needs more debugging to understand where is the root cause of it.

PR #394, opened 2022-12-02
Error expansion seems to be missed in clevis-luks-report

sarroutbi:
Fixes: #393


sarroutbi:
I have a doubt regarding this change, as when using "sss" there are multiple "configurations" that should be tried. If, for example, you have two "tang" servers, one of them down, should it be "ignored" to try to configure the second one?
What's your opinion, @sergio-correia , is this intentionally? If so, please let me know to close this PR.

PR #399, opened 2023-01-22
Add fido2 pin

olastor:
Implementation proposal for #332 

- adds a new pin for encrypting using a fido2 token, e.g., yubikey.
- tokens must support the hmac-extension for it to work
- requires [libfido2](https://developers.yubico.com/libfido2/) and the commands `fido2-cred` and `fido2-assert`
- creates a new credential by default, but it's also possible to pass an existing credential id (see docs)
- chosen "alg" for the JWK that uses the secret derived via the token is ~"PBES2-HS512+A256KW"~ AES256GCM
- I am not familiar with meson, so the libfido2 dependency is currently not linked

Example usage:

```
echo test | ./clevis-encrypt-fido2 '{"up": false, "uv": false}' | ./clevis-decrypt-fido2
```

Would appreciate any reviews, especially by people familiar with FIDO2/WebAuthn/CTAP to make sure the design is reasonably secure. Afaik, creating a non-discoverable credential and storing it as public metadata is similar as systemd-cryptenroll implements it.

### DIFF
--- a/src/pins/fido2/clevis-decrypt-fido2
+++ b/src/pins/fido2/clevis-decrypt-fido2
@@ -18,7 +18,7 @@
 
 set -eu
 
-read -d . hdr64
+read -r -d . hdr64
 if ! hdr="$(jose fmt --quote="$hdr64" --string --b64load --object --output=-)" ; then
     echo 'JWE header corrupt' >&2
     exit 1
@@ -28,28 +28,28 @@ if [ "$(jose fmt --json="$hdr" --get clevis --get pin --unquote=-)" != 'fido2' ]
     exit 1
 fi
 if ! hmac_salt="$(jose fmt --json="$hdr" --get clevis --get fido2 --get hmac_salt --unquote=-)" ; then
-    echo 'JWE missing 'hmac_salt' header parameter!' >&2
+    echo "JWE missing 'hmac_salt' header parameter!" >&2
     exit 1
 fi
 
 if ! rp_id="$(jose fmt --json="$hdr" --get clevis --get fido2 --get rp_id --unquote=-)" ; then
-    echo 'JWE missing 'rp_id' header parameter!' >&2
+    echo "JWE missing 'rp_id' header parameter!" >&2
     exit 1
 fi
 if ! cred_id="$(jose fmt --json="$hdr" --get clevis --get fido2 --get cred_id --unquote=-)" ; then
-    echo 'JWE missing 'cred_id' header parameter!' >&2
+    echo "JWE missing 'cred_id' header parameter!" >&2
     exit 1
 fi
 if ! uv="$(jose fmt --json="$hdr" --get clevis --get fido2 --get uv --unquote=-)" ; then
-    echo 'JWE missing 'uv' header parameter!' >&2
+    echo "JWE missing 'uv' header parameter!" >&2
     exit 1
 fi
 if ! up="$(jose fmt --json="$hdr" --get clevis --get fido2 --get up --unquote=-)" ; then
-    echo 'JWE missing 'up' header parameter!' >&2
+    echo "JWE missing 'up' header parameter!" >&2
     exit 1
 fi
 if ! pin="$(jose fmt --json="$hdr" --get clevis --get fido2 --get pin --unquote=-)" ; then
-    echo 'JWE missing 'pin' header parameter!' >&2
+    echo "JWE missing 'pin' header parameter!" >&2
     exit 1
 fi
 
--- a/src/pins/fido2/clevis-decrypt-fido2
+++ b/src/pins/fido2/clevis-decrypt-fido2
@@ -71,10 +71,10 @@ client_hash="$(dd if=/dev/urandom bs=1 count=32 status=none | base64 -w0)"
 
 hmac="$(printf '%s\n%s\n%s\n%s\n' "${client_hash}" "${rp_id}" "${cred_id}" "${hmac_salt}" | \
   fido2-assert -G -t "uv=${uv}" -t "up=${up}" -t "pin=${pin}" -h "${fido2_token}" | \
-  head -n5 | tail -n1 | jose b64 enc -I -)"
+  head -n5 | tail -n1 | base64 -d | jose b64 enc -I -)"
 
 # use the secret in a key wrapping key
-jwk='{"alg":"PBES2-HS512+A256KW", "kty":"oct"}'
+jwk='{"alg":"A256GCM", "kty":"oct"}'
 jwk="$(jose fmt -j "${jwk}" -q "${hmac}" -s k -Uo-)"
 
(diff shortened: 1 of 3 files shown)


sarroutbi:
Hello @olastor . Thanks for the PR. I have two doubts regarding it:

1. Why did you not include a man page for the decryption part?
2. What about the error dumped by the code scanning tool?:
```
jwe='{"protected":{"enc":"A256GCM","clevis":{"pin":"fido2","fido2":{}}}}'
jwe="$(jose fmt -j "$jwe" -g protected -q "$kid" -s kid -UUo-)"
```
Why is this *kid* variabled used, if it has not been previously assigned to any value?

src/pins/fido2/clevis-decrypt-fido2
```diff
@@ -0,0 +1,80 @@
...
+if ! hdr="$(jose fmt --quote="$hdr64" --string --b64load --object --output=-)" ; then
+    echo 'JWE header corrupt' >&2
+    exit 1
+fi
+if [ "$(jose fmt --json="$hdr" --get clevis --get pin --unquote=-)" != 'fido2' ] ; then
+    echo 'JWE pin mismatch!' >&2
+    exit 1
+fi
+if ! hmac_salt="$(jose fmt --json="$hdr" --get clevis --get fido2 --get hmac_salt --unquote=-)" ; then
+    echo 'JWE missing 'hmac_salt' header parameter!' >&2
```
sarroutbi, line 31: As you are using simple quotes, you could do:
echo "JWE .... 'hmac_salt' .... ! "

src/pins/fido2/clevis-decrypt-fido2
```diff
@@ -0,0 +1,80 @@
...
+if ! rp_id="$(jose fmt --json="$hdr" --get clevis --get fido2 --get rp_id --unquote=-)" ; then
+    echo 'JWE missing 'rp_id' header parameter!' >&2
+    exit 1
+fi
+if ! cred_id="$(jose fmt --json="$hdr" --get clevis --get fido2 --get cred_id --unquote=-)" ; then
+    echo 'JWE missing 'cred_id' header parameter!' >&2
+    exit 1
+fi
+if ! uv="$(jose fmt --json="$hdr" --get clevis --get fido2 --get uv --unquote=-)" ; then
+    echo 'JWE missing 'uv' header parameter!' >&2
```
sarroutbi, line 44: As you are using simple quotes, you could do:
echo "JWE .... 'uv' .... ! "

This is repeated in different lines, so please, review them if possible.

Shellcheck complains about it this way:
In clevis-decrypt-fido2 line 44:
    echo 'JWE missing 'uv' header parameter!' >&2
                       ^-- SC2026 (info): This word is outside of quotes. Did you intend to 'nest '"'single quotes'"' instead'?

olastor: Right, thanks for the hint. I fixed that and the above comment in the latest commit :ok_hand:

olastor:
> Hello @olastor . Thanks for the PR. Why did you not include a man page for the decryption part?

Hi @sarroutbi , thank you for taking a look at the PR! The man [clevis decrypt man page](https://github.com/latchset/clevis/blob/master/src/clevis-decrypt.1.adoc) seems to be "pin-agnostic". None of the other pins has one, so I think this pin wouldn't need one, either. Only for encrypting, there are specific options that need to be documented.

src/pins/fido2/clevis-encrypt-fido2
```diff
@@ -0,0 +1,109 @@
...
+# retrieve the hmac result which will be the password to use for key wrapping a CEK.
+hmac="$(generate_hmac "${fido2_token}" "${rp_id}" "${cred_id}" "${hmac_salt}")"
+
+if [ -z "${hmac}" ]; then
+  echo "Error: could not generate key."
+  exit 1
+fi
+
+# use the secret in a key wrapping key
+jwk='{"kty":"oct", "alg":"PBES2-HS512+A256KW"}'
```
olastor, line 97: If anyone is reviewing this, I'd like to know your opinion on my choice of this "alg" here (if any).

sarroutbi:
> > Hello @olastor . Thanks for the PR. Why did you not include a man page for the decryption part?
> 
> Hi @sarroutbi , thank you for taking a look at the PR! The man [clevis decrypt man page](https://github.com/latchset/clevis/blob/master/src/clevis-decrypt.1.adoc) seems to be "pin-agnostic". None of the other pins has one, so I think this pin wouldn't need one, either. Only for encrypting, there are specific options that need to be documented.

Hello @olastor. Thanks for clarifying. You are completely right. I did not realize it was "pin-agnostic"

sarroutbi (commented):
In general, change LGTM. If possible, it would be nice to add a unit test to check FIDO2 pin works appropriately

src/pins/fido2/clevis-encrypt-fido2
sarroutbi: Sorry, I don't have a strong opinion regarding algorithm selection.

sarroutbi (approved):
Changes LGTM

sarroutbi:
@sergio-correia : can you please provide your feedback regarding this PR when possible?

src/pins/fido2/clevis-encrypt-fido2
olastor: **Update**: I changed this to AES256GCM. The thinking behind the exotic first choice of algorithm here was that it would be better to additionally do some key derivation from the hmac secret retrieved from the fido2 token, and also because I experienced some wrapping errors when using the raw value returned. Coming back to this, I changed my mind. I don't think an additional key derivation might be needed here as the hmac over the random salt should be sufficient (but I am happy to be proven wrong about it). The wrapping errors I experienced with the raw value were likely due to libfido2 returning values encoded in base64, and I probably didn't get the conversion to base64url right, which should now be fixed. So now the hmac output is used as a simple `AES256GCM` type key.

olastor:
@sarroutbi @sergio-correia It's been a while since I opened this PR and I am wondering how big the chances of merging this in the future are? I am willing (given I find the time) to look into how to create tests or improve the docs/code if necessary, or add a notice about it being an experimental pin in the beginning. There are for example some possible UX improvements like waiting for the key to be inserted or more informative messages for the user.  However, if this PR stays "stale", then I will probably move forward to maintaining this pin in a separate private repo instead.

sarroutbi:
Hello @olastor. As I explained before, I prefer to have a double check on this PR, taking into account this is a big change, I would like a second opinion. @sergio-correia : can you please provide your feedback regarding this PR when possible?

olastor:
I am going to maintain this pin in a separate repository (https://github.com/olastor/clevis-pin-fido2) for now. Maybe I'll reopen the PR in the future if this project has more active maintainers.

PR #402, opened 2023-01-31
Include --netrc-file when calling curl

sarroutbi:
Resolves: #401

### DIFF
--- a/src/pins/tang/clevis-decrypt-tang
+++ b/src/pins/tang/clevis-decrypt-tang
@@ -101,7 +101,18 @@ xfr="$(jose jwk exc -i '{"alg":"ECMR"}' -l- -r- <<< "$clt$eph")"
 
 rec_url="$url/rec/$kid"
 ct="Content-Type: application/jwk+json"
-if ! rep="$(curl -sfg -X POST -H "$ct" --data-binary @- "$rec_url" <<< "$xfr")"; then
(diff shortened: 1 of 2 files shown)


sergio-correia (changes requested):
Let's discuss the enhancement before diving into the code, but looking into the PR, I am not a fan of that harcoded `/root/.netrc` -- is this feature only intended to be used by root?

sarroutbi:
> Let's discuss the enhancement before diving into the code, but looking into the PR, I am not a fan of that harcoded `/root/.netrc` -- is this feature only intended to be used by root?

Well, I don't see many cases where binding is required as "non-root" user, as it normally requires access to a device. But yes, in an ideal case, this should be passed as external argument, I agree with that. I'll try to figure out how to do this, as there is no "clevis configuration file" where this could be configured

sarroutbi:
> Let's discuss the enhancement before diving into the code, but looking into the PR, I am not a fan of that harcoded `/root/.netrc` -- is this feature only intended to be used by root?

As this is used for encryption and decryption (also at boot time), I suggest using a configuration file under /etc. Change contemplates the possibility of this file not to exist, leaving execution intact (file can not exist, and everything will continue to work similarly)

sarroutbi:
Unnecessary at this moment

PR #417, opened 2023-05-23
Fix README.md to include tang https configuration (#175)

MaxiStarling56:


README.md
```diff
@@ -139,7 +139,9 @@ Upon successful completion of this binding process, the disk can be unlocked
 using one of the provided unlockers.
 
 #### Network based unlocking
-If you want to use network based unlocking you will need to specify `rd.neednet=1` as kernel argument or use `--hostonly-cmdline` when creating dracut.
+If you want to use network based unlocking you will need to specify `rd.neednet=1` as kernel argument or use `--hostonly-cmdline` when creating with dracut.
+
+If you're using **tang** as pin and the tang server is using tls (Example: `'{"url": "https://tang.remote"}'`), the folder `/etc/ssl` should be included in the initramfs image, `--include /etc/ssl /etc/ssl --force` when creating with dracut.
```
sarroutbi, line 144: Normally, documentation refers to Tang (not tang) for Tang server. Also, use capital letters (TLS instead of tls) for acronyms.

sarroutbi (changes requested):

README.md
MaxiStarling56: So, it should be like the following?
```Markdown
If you're using **Tang** with TLS (Example: `'{"url": "https://tang.remote"}'`), the folder `/etc/ssl` should be included in the initramfs image, `--include /etc/ssl /etc/ssl --force` when creating with dracut.
```

Thanks for the feedback

sergio-correia: It might be better to change the dracut/initramfs-tools modules to include the required certificates, so that it would work with HTTPS out of the box.

MaxiStarling56: > It might be better to change the dracut/initramfs-tools modules to include the required certificates, so that it would work with HTTPS out of the box.

Tried that but wasn't able to, I'm not very good on linux but will try again, anyway I think that while there isn't a better solution the readme shoud at least inform the users  that there is this workaround.

README.md
```diff
@@ -139,7 +139,9 @@ Upon successful completion of this binding process, the disk can be unlocked
 using one of the provided unlockers.
 
 #### Network based unlocking
-If you want to use network based unlocking you will need to specify `rd.neednet=1` as kernel argument or use `--hostonly-cmdline` when creating dracut.
+If you want to use network based unlocking you will need to specify `rd.neednet=1` as kernel argument or use `--hostonly-cmdline` when creating with dracut.
+
+If you're using **Tang** with TLS (Example: `'{"url": "https://tang.remote"}'`), the folder `/etc/ssl` should be included in the initramfs image, `--include /etc/ssl /etc/ssl --force` when creating with dracut.
```
sarroutbi, line 144: When specifying this:
```
--include /etc/ssl /etc/ssl --force
```
Is it necessary to provide `/etc/ssl` twice?

MaxiStarling56: Yes, the first is "_where to copy the files from?_", the second is "_where to put it in the builded initramfs image?_", more info [here](https://linuxconfig.org/how-to-build-an-initramfs-using-dracut-on-linux), in the section "Including extra files in the initramfs with the --include option"

sarroutbi (approved):
Change LGTM

sergio-correia (approved):

PR #421, opened 2023-06-01
Fix dnf.conf in case a section is missed

sarroutbi:
Resolves: #420

### DIFF
--- a/.github/workflows/install-dependencies
+++ b/.github/workflows/install-dependencies
@@ -30,7 +30,16 @@ debian:*|ubuntu:*)
     ;;
 
 fedora:*)
+    grep -e '\[[a-z]*\]' /etc/dnf/dnf.conf ||
+    {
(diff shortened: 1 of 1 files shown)


sarroutbi:
This is necessary to fix latest issues on Fedora Rawhide and DNF, which expects a section in dnf.conf, while distro is not adding it:
https://github.com/latchset/clevis/actions/runs/5142022163/jobs/9255230165

sarroutbi:
This seems related to a recent issue by mixing dnf3 and dnf5 in Fedora Rawhide:
https://www.spinics.net/lists/fedora-testing/msg146969.html

Let's wait a pair of days until merging it

sarroutbi:
This approach won't be followed. In case it is needed, quay.io version of Fedora Rawhide will be used, as this version is working appropriately

PR #449, opened 2024-01-28
Fix killing of child process of clevisloop

### LINKED ISSUE #448: Not working killing of child process of clevisloop
EDIT: Updated, I initially thought this is about killing of parent process, but it is actually about a child process

I am migrating my script with TPM1.2 support to Clevis and found this strange line of code in Clevis scripts:

https://github.com/latchset/clevis/blob/fee1db33c8e33a6ec973bf6124fb5dc1497e81b1/src/initramfs-tools/scripts/local-bottom/clevis.in#L37

It looks like the intention was to kill the child process of the `clevisloop` itself. But it uses fragile `ps -l`, which has different output on normal system and different versions of BusyBox - there is different set of columns in several versions.

I am under Debian 12 (Bookworm) and added some debug output in clevis local-down script and found that it really cannot work like this.

The `clevisloop` PID was 275:

 1. `ps -l` output shows the following (see header - different set of columns):

```text
PID   USER     COMMAND
    1 0        {init} /bin/sh /init
  ...
  275 0        {clevis} /bin/bash /scripts/local-top/clevis
  ...
  436 0        sleep 5
  ...
  596 0        {clevis} /bin/sh /scripts/local-bottom/clevis
  ...
  598 0        ps -l
```

Really, the BusyBox's `ps` in Debian does not do much:

```
#> busybox ps --help
BusyBox v1.35.0 (Debian 1:1.35.0-4+b3) multi-call binary.

Usage: ps [-o COL1,COL2=HEADER] [-T]

Show list of processes

        -o COL1,COL2=HEADER     Select columns for display
        -T                      Show threads
```

So I used more portable `ps -o ppid,pid,comm` for output (not killing) and found the following:

```text
PPID  PID   COMMAND
    0     1 init
  ...
    1   275 clevis
  ...
  275   436 sleep
  ...
    1   596 clevis
  596   599 ps
```

It should be possible to find the child process with `ps -o ppid,pid`.

oldium:
The usage of `ps -l` is fragile and does not always return the expected set of columns. It differs even between several BusyBox versions and operating systems. Use `ps -o pid,ppid` to get the expected set of columns instead.

Also kill parent first before its children to prevent spawning another command from clevisloop itself.

Fixes #448

### DIFF
--- a/src/initramfs-tools/scripts/local-bottom/clevis.in
+++ b/src/initramfs-tools/scripts/local-bottom/clevis.in
@@ -34,8 +34,10 @@ esac
 [ -s /run/clevis.pid ] || exit 0
 
 pid=$(cat /run/clevis.pid)
-ps -l | awk -v pid="$pid" '$4==pid { system("kill " $3) }'
-kill "$pid"
+child_pids=$(ps -o pid,ppid | awk -v pid="$pid" '$2==pid { print $1 }')
+for kill_pid in $pid $child_pids; do
+	kill "$kill_pid"
+done
 
 # Not really worried about downing extra interfaces: they will come up
 # during the actual boot. Might make this configurable later if needed.


sergio-correia (approved):

sarroutbi (approved):
Hello. Thanks for your PR. Changes LGTM

PR #453, opened 2024-03-05
Miscelaneous fixes

sergio-correia:
- luks/udisks2: explicitly NULL-terminate buffer
- pins/sss: intialize variable before use


sarroutbi (commented):
Change LGTM

sarroutbi (approved):

PR #464, opened 2024-05-09
udisks2: misc fixes

sergio-correia:
- udisks2: check if variables are NULL before calling unref
- udisks2: check for EINTR when reading in recover_key()

### DIFF
--- a/src/luks/udisks2/clevis-luks-udisks2.c
+++ b/src/luks/udisks2/clevis-luks-udisks2.c
@@ -301,12 +301,12 @@ on_signal(int sig)
     safeclose(&pair[0]);
 }
 
-static ssize_t
-recover_key(const pkt_t *jwe, char *out, size_t max, uid_t uid, gid_t gid)
+static uint32_t
+recover_key(const pkt_t *jwe, char *out, int32_t max, uid_t uid, gid_t gid)
 {
     int push[2] = { -1, -1 };
     int pull[2] = { -1, -1 };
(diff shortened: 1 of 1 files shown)


src/luks/udisks2/clevis-luks-udisks2.c
```diff
@@ -264,8 +264,10 @@ child_main(int sock)
 
 error:
     g_list_free_full(ctx.lst, g_free);
```
sarroutbi, line 266: In this case, we might protect also the call to `g_list_free_full`:
```
  if(ctx.lst) {
    g_list_free_full(ctx.lst, g_free);
  }
```
I doubt g_free is NULL protected

sarroutbi (commented):
Minor comment. Change LGTM

src/luks/udisks2/clevis-luks-udisks2.c
sergio-correia: They usually are: https://docs.gtk.org/glib/func.free.html
But I can add a check there too, if you want.

sarroutbi: > They usually are: https://docs.gtk.org/glib/func.free.html But I can add a check there too, if you want.

Sorry. Not necessary to add redundant checks. I thought it would work as `free`. Change LGTM

sarroutbi (approved):
Change LGTM

PR #466, opened 2024-05-28
Ensure dnf builddep is installed in Fedora

sarroutbi:
Resolves: #465


.github/workflows/install-dependencies
```diff
@@ -34,6 +34,7 @@ debian:*|ubuntu:*)
     dnf -y clean all
     dnf -y --setopt=deltarpm=0 update
     dnf -y install dnf-utils jq socat cryptsetup keyutils cracklib-dicts lsof
+    type dnf5 && dnf5 -y install dnf5-command\(builddep\) || dnf -y install dnf-command\(builddep\)
```
sergio-correia, line 37: Looks good, but please use `command -v dnf5` instead of `type` here.

sergio-correia (approved):

PR #470, opened 2024-07-03
CentOS Stream 8 test build fix

oldium:
This fixes build on CentOS Stream 8 after mirrorlist.centos.org vanished.


oldium:
Alternative would be to remove CentOS Stream 8 support, it is EOL already.

oldium:
Not necessary anymore, fixed in master already.

PR #471, opened 2024-07-12
[EXPERIMENTAL] Add PKCS#11 pin basic functionality

sarroutbi:
Include experimental PKCS#11 pin to allow disk unlocking with PKCS#11 devices. For extended information on installation and configuration, check README.md
    
Co-authored-by: Sergio Correia <scorreia@redhat.com>


src/pins/pkcs11/clevis-encrypt-pkcs11
```diff
@@ -0,0 +1,135 @@
...
+    module_opt=""
+else
+    module_opt=" --module ${module_path}"
+fi
+
+if ! slot=$(clevis_get_slot_from_uri "${uri}"); then
+    slot_opt=""
+else
+    slot_opt=" --slot-index ${slot}"
+fi
```
Jakuje, line 81: Note, that the slot numbers are not guaranteed to be stable over invocations so this should be at least discouraged in the documentation that this might be problematic. OTOH the RFC says that it might be useful so I would not remove it, but add there a warning describing the issue similarly as in the RFC:

https://datatracker.ietf.org/doc/html/rfc7512#page-9

src/pins/pkcs11/clevis-encrypt-pkcs11.1.adoc
```diff
@@ -0,0 +1,76 @@
...
+* *uri*  (string) :
+  The PKCS#11 URI to use (REQUIRED)
+
+* *mechanism*  (string) :
+  Mechanism to be used when working with pkcs11-tool.
+  It must be supported by pkcs11-tool. Examples of supported mechanisms are:
+  - *RSA-PKCS-OAEP* (default one)
+  - *RSA-PKCS*
+  - *SHA256*
+  - *SHA512*
```
Jakuje, line 68: These are not a RSA encryption mechanisms so please remove them.

src/pins/pkcs11/clevis-encrypt-pkcs11.1.adoc
```diff
@@ -0,0 +1,76 @@
...
+
+For a complete list of supported mechanisms, execute next command:
+
+  $ pkcs11-tool -M
```
Jakuje, line 72: I would mention here that we need mechanisms that support encryption and decryption (the flags are printed in this output).

src/pins/pkcs11/clevis-decrypt-pkcs11
```diff
@@ -0,0 +1,134 @@
...
+
+if ! jwk="$(pkcs11-tool --login --decrypt --input-file ${ENC} \
+            -p ${PIN_value} --module ${module_path} ${mechanism_option} 2>${ERR})" \
+            || [ -z "${jwk}" ]; then
+    cat "${ERR}" >&2
+    echo "Unable to decrypt the JWK" >&2
+    # Check if it is an issue with PIN (to dump error appropriately)
+    if ! pkcs11-tool -pkcs11-tool --login --test -p "${PIN_value}"-login --test \
+                     -p "${PIN_value}" --module "${module_path}" \
+                     2>"${ERR}" >/dev/null; then
```
Jakuje, line 124: I am not sure this is a good idea. The cards have limited amount of wrong pin inputs before the card locks. This way you try twice the same PIN and if it is wrong which basically drops this to half. The second pin will always be wrong if I read the value `"${PIN_value}"-login`.

I hope there should be a better way to distinguish the error from one call. We should consider looking into using the pkcs11-provider or directly PKCS#11 API.

Moreover, you have the `--test` argument in this command-line twice, but the `--test` runs some tests on all keys, which is certainly not something you want to do in the production.

README.md
```diff
@@ -77,6 +77,102 @@ $ echo hi | clevis encrypt tpm2 '{}' > hi.jwe
...
+
+In the first phase of development, Clevis will be used in top of OpenSC to provide PKCS#11 functionality.
+OpenSC, and, in particular, `pkcs11-tool`, provides an option to indicate the mechanism to use for decryption.
+For testing purposes, some libraries, such as [SoftHSM](https://www.opendnssec.org/softhsm)), don't work with default `pkcs11-tool` mechanism,
+so it is  required to provide a particular mechanism to use. For this reason, Clevis can be provided with
+the mechanism to use, in case the default one, `RSA-PKCS-OAEP`, is not valid:
+
+```
+$ clevis luks bind -d /dev/sda1 pkcs11 '{"uri": "pkcs11:", "mechanism":"RSA-PKCS"}'
+```
```
Jakuje, line 174: I would put here a note that the RSA-PKCS mechanism (PKCS#1.5 padding for encryption) is not considered secure these days as discussed before and it is mostly provided for compatiblity, but not recommended:
 https://people.redhat.com/~hkario/marvin/

src/pins/pkcs11/clevis-decrypt-pkcs11
sarroutbi: @Jakuje : Thanks for your suggestion. What do you think about removing this check and just assume the cause is an invalid PIN? In the end, this will be temporary until we include pkcs11-provider

Jakuje: For now, probably removing it would be best and assuming wrong PIN (that should not break anything). And maybe keeping some note that this need some better handling in the future.

README.md
```diff
@@ -77,6 +77,216 @@ $ echo hi | clevis encrypt tpm2 '{}' > hi.jwe
...
+$ clevis luks bind -d /dev/sda1 pkcs11 '{"uri": "pkcs11:model=PKCS%2315%20emulated;manufacturer=piv_II;
+serial=0a35ba26b062b9c5;token=clevis;id=%02;object=Encryption%20Key"}'
+```
+
+##### Configuration to bind Clevis to the first PKCS#11 device found
+An additional option is to provide Clevis a configuration so that the first PKCS#11 device found by Clevis is bound. To do so, an empty URI can be provided as shown below:
+
+```
+$ clevis luks bind -d /dev/sda1 pkcs11 '{"uri":, "pkcs11:"}'
```
sergio-correia, line 125: There is a typo there. the comma makes the config not a valid JSON.

Also, what about making this the default, in case no `uri` is provided?

sarroutbi: Do you mean allowing something like this?:

`$ clevis luks bind -d /dev/sda1 pkcs11 '{}'`

sarroutbi: OK. I will fix it and will allow "non uri" as "uri":"pkcs11:"

README.md
```diff
@@ -77,6 +77,239 @@ $ echo hi | clevis encrypt tpm2 '{}' > hi.jwe
...
+
+4 - Enable clevis-luks-pkcs11-askpass.socket unit:
+
+```
+$ sudo systemctl enable --now clevis-luks-pkcs11-askpass.socket
+```
+
+5 - /etc/crypttab configuration:
+
+As described in [Integration with systemd](#integration-with-systemd) section, `crypttab` must be configured so that systemd uses an AF\_UNIX socket to wait for the keyphrase that will unlock the disk and not to prompt it through the console.
```
sergio-correia, line 286: Where is this "integration with systemd" section?

sarroutbi: > Where is this "integration with systemd" section?

I will remove it. Such a section does not exist

sergio-correia (approved):
Good job, thanks. I think this look good for an initial version of this.

PR #481, opened 2024-09-30
PKCS#11 pin: fix dracut for unconfigured device

sarroutbi:
Hot fix to sort out execution of dracut when no PKCS#11 device exists, no configuration has been performed in /etc/crypttab but, even in this case, empty device is prompted (when it should not)


sergio-correia (approved):
Thanks, LGTM

PR #484, opened 2024-10-05
tpm2: fix dangling loaded session after tpm2_createpolicy

oldium:
Clevis cleans the context regurarly from transient objects by invoking `tpm2_flushcontext -t`, but there is one case which Clevis omits. The `tpm2_createpolicy` command loads new session and this needs to be cleaned by a call to `tpm2_flushcontext -l`, see https://github.com/tpm2-software/tpm2-tools/issues/3427.

This is usually not an issue when resource manager is used (i.e. with `/dev/tpmrm0` device), but when raw TPM device `/dev/tpm0` or software TPM (swtpm) without resource manager is used, the usage of clevis (multiple calls are required) ends with the following error:

```
WARNING:esys:src/tss2-esys/api/Esys_StartAuthSession.c:390:Esys_StartAuthSession_Finish() Received TPM Error
ERROR:esys:src/tss2-esys/api/Esys_StartAuthSession.c:136:Esys_StartAuthSession() Esys Finish ErrorCode (0x00000903)
ERROR: Esys_StartAuthSession(0x903) - tpm:warn(2.0): out of memory for session contexts
ERROR: Unable to run tpm2_create
```

### DIFF
--- a/src/pins/tpm2/clevis-encrypt-tpm2
+++ b/src/pins/tpm2/clevis-encrypt-tpm2
@@ -223,6 +223,7 @@ if [ -n "$pcr_ids" ]; then
         exit 1
     fi
     tpm2_flushcontext -t
+    tpm2_flushcontext -l
 
     policy_options+=(-L "$TMP/pcr.policy")
 else


oldium:
It seems they will be fixing the behaviour of `tpm2-createpolicy` (see https://github.com/tpm2-software/tpm2-tools/issues/3427#issuecomment-2395079701), but in order to have this working with older versions, I would keep the workaround with `tpm2_flushcontext -l` for now.

oldium:
Reproduction with `clevis` is actually simple - either use swtpm (I discovered this with swtpm originally) or raw TPM device (`/dev/tpm0`) and run tpm2 pin tests. Or the following:

```console
# export TPM2TOOLS_TCTI=device:/dev/tpm0
# tpm2_getcap pcrs
selected-pcrs:
  - sha1: [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23 ]
  - sha256: [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23 ]
# echo "test" | clevis encrypt tpm2 '{"pcr_ids":"23"}' | clevis decrypt
test
# echo "test" | clevis encrypt tpm2 '{"pcr_ids":"23"}' | clevis decrypt
WARNING:esys:src/tss2-esys/api/Esys_StartAuthSession.c:390:Esys_StartAuthSession_Finish() Received TPM Error
ERROR:esys:src/tss2-esys/api/Esys_StartAuthSession.c:136:Esys_StartAuthSession() Esys Finish ErrorCode (0x00000903)
ERROR: Esys_StartAuthSession(0x903) - tpm:warn(2.0): out of memory for session contexts
ERROR: Unable to run tpm2_create
Creating TPM2 object for jwk failed!
```

sarroutbi (commented):
Thanks for your PR, @oldium . Sorry for my ignorance but ... is it necessary to call it twice? Is not a call to this command valid?:
```bash
tpm2_flushcontext -t -l
```

oldium:
> Thanks for your PR, @oldium . Sorry for my ignorance but ... is it necessary to call it twice? Is not a call to this command valid?:
> 
> ```shell
> tpm2_flushcontext -t -l
> ```

Your question is right on target, I have checked exactly that. 😁 This is supported since `tpm2-tools` 5.3, but unfortunately Debian 11 ships `tpm2-tools` 5.0...

oldium:
But meson could check for that, safe command to execute is:

```
TPM2TOOLS_TCTI=none tpm2_flushcontext -t -l
```

When it reports `ERROR: Options -t, -l and -s are mutually exclusive`, the flushing combo is not supported. And it is forwards compatible, so any update to tpm2-tools in future cannot break the current clevis release.

The correct command/sequence can be generated during build (with `configure_file`). What do you think?

sarroutbi:
> @sarroutbi But meson could check for that, safe command to execute is:
> 
> ```
> TPM2TOOLS_TCTI=none tpm2_flushcontext -t -l
> ```
> 
> When it reports `ERROR: Options -t, -l and -s are mutually exclusive`, the flushing combo is not supported. And it is forwards compatible, so any update to tpm2-tools in future cannot break the current clevis release.
> 
> The correct command/sequence can be generated during build (with `configure_file`). What do you think?

Thanks for clarifying. In that case, I would keep it as it is now.

sarroutbi (approved):
LGTM

PR #486, opened 2024-10-05
pkcs11: provide key file format to openssl

oldium:
When built under Debian-11, the following error appears:

```
++ echo secret
++ clevis decrypt
++ clevis encrypt pkcs11 '{"uri":"pkcs11:module-path=/usr/lib/softhsm/libsofthsm2.so?pin-value=123456","mechanism":"RSA-PKCS"}'
unable to load Public Key
Unable to encrypt JWK with PKCS#11 public key
```

Fix this by telling the key file format (DER) to OpenSSL.


sarroutbi (approved):
Change LGTM

PR #487, opened 2024-10-05
systemd: search for systemd-reply-password in dir specified by systemd

oldium:
This increases build compatibility between systems, like Debian 11 and Debian 12.


sarroutbi (approved):
LGTM

PR #489, opened 2024-10-05
tang: clean-up unused variable, omit reuseaddr

oldium:
The `port` argument is not used, so remove it. The `reuseaddr` flag is useless, we use fresh new port, we are not reusing any.

The second `reuseaddr` is cleaned-up in #488.


sarroutbi (approved):
Change LGTM

PR #490, opened 2024-10-05
tpm2: use first PCR algorithm bank supported by TPM as default

oldium:
The default is `sha1`, which is not always supported (it is legacy and optional for implementation). Make this more future-proof and use the first bank with non-empty set of PCRs, which is returned from TPM by `tpm2_getcap pcrs`.

The swtpm by default does not create sha1 bank, so this fixes usage with swtpm.


oldium:
Small update, simplified exit logic in `awk`.

sergio-correia (approved):
Thanks

PR #496, opened 2024-10-31
initramfs: fix killing child PIDs

oldium:
Ubuntu's initramfs Busybox does not support `ps -o pid,ppid`, so fallback to unpredictable `ps -l`. In any case use `awk` to find the position of PID and PPID columns and dump appropriate child PIDs. This should be hopefully bullet-proof.

Also clean-up leftover PID file.

Fixes issue [found](https://github.com/latchset/clevis/pull/462#issuecomment-2439735757) when running under Ubuntu's version of Busybox.


sergio-correia (approved):
Thanks, LGTM

PR #500, opened 2024-11-07
Groom references to deprecated tangd-update script

sarroutbi:
Resolves: #497


src/pins/tang/tests/meson.build
```diff
@@ -43,10 +42,6 @@ if tang.found()
   tang_data.set('TANGD', tang.path())
 endif
 
-if updt.found()
```
sergio-correia, line 46: Check also where `updt` is defined and remove it, please. Search for `tangd-update` to find it quicker.

sergio-correia (approved):

PR #493, opened 2024-10-12
luks: make sure cryptsetup.target is installed

LaszloGombos:
dracut [v103 ](https://github.com/dracut-ng/dracut-ng/releases/tag/103) introduced systemd-cryptsetup.

dracut [v104](https://github.com/dracut-ng/dracut-ng/releases/tag/104) requires systemd-cryptsetup for cryptsetup.target
(https://github.com/dracut-ng/dracut-ng/commit/ad520855113d99d7551a7ab6c19870736f72cbc4.

CC @BtbN @cornelicorn

PR #501, opened 2024-11-09
dracut: fix running with pre-v103 Dracut

oldium:
Dracut v103+ moved SystemD cryptsetup code into a separate module, so detect it and when found, use it.

Clevis units do not depend on `cryptsetup.target`, so do not install it. The target `cryptsetup.target` is installed by both [90systemd-cryptsetup][1] in v103+ and [01systemd-cryptsetup][2] in v105+, so there is no need to install it by Clevis.

Fixes: #493

Note: Tested with Dracut v102 (Fedora 40) and v103 (Fedora Rawhide).

[1]: https://github.com/dracut-ng/dracut-ng/blob/3fce598fb45aa5618cdf885eb48cf327104ffcb8/modules.d/90systemd-cryptsetup/module-setup.sh#L53
[2]: https://github.com/dracut-ng/dracut-ng/blob/a2669c447bd5e0406f55efcc8c9c58be89458b08/modules.d/01systemd-cryptsetup/module-setup.sh#L54

### DIFF
--- a/src/luks/dracut/clevis/module-setup.sh.in
+++ b/src/luks/dracut/clevis/module-setup.sh.in
@@ -21,7 +21,13 @@
 depends() {
     local __depends=crypt
     if dracut_module_included "systemd"; then
-        __depends=$(printf '%s systemd-cryptsetup' "${__depends}")
+        # Dracut v103 introduced a separate systemd-cryptsetup module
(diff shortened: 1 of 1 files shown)


oldium:
Rebased to laster master

sergio-correia (approved):
Thanks!

PR #503, opened 2024-11-25
Include tpm2_getcap as required tpm2 pin binary

sarroutbi:


jlebon:
For context, Ignition traditionally has pulled in all the Clevis/TPM-related executables into the initramfs related to _binding_ since that's understandably not covered by the dracut modules here:

https://github.com/coreos/ignition/blob/e54b43f7ca60ac1a7ff7ff078359ad72458dda00/dracut/30ignition/module-setup.sh#L49-L61

If that's the case here (i.e. that `tpm2_getcap` is only required at binding time), then it probably makes more sense to leave this in Ignition only. If it's also required for unlocking, then it should live here.

Relatedly, it'd be great if we could move that Ignition list into a separate Clevis dracut module here that would be easier to keep up to date. Though probably to do this well we'd want to add CI testing as well. This assumes there's interest beyond Ignition in doing binding from the initramfs.

sarroutbi:
> For context, Ignition traditionally has pulled in all the Clevis/TPM-related executables into the initramfs related to _binding_ since that's understandably not covered by the dracut modules here:
> 
> https://github.com/coreos/ignition/blob/e54b43f7ca60ac1a7ff7ff078359ad72458dda00/dracut/30ignition/module-setup.sh#L49-L61
> 
Thanks for the explanation

> If that's the case here (i.e. that `tpm2_getcap` is only required at binding time), then it probably makes more sense to leave this in Ignition only. If it's also required for unlocking, then it should live here.
> 
In case clevis-pin-tpm2 is not used, `tpm2_getcap` is needed for unlocking

> Relatedly, it'd be great if we could move that Ignition list into a separate Clevis dracut module here that would be easier to keep up to date. Though probably to do this well we'd want to add CI testing as well. This assumes there's interest beyond Ignition in doing binding from the initramfs.
We would welcome a PR for that if you find it interesting

jlebon (approved):
In that case, LGTM! I noticed you backported this to Fedora already. Can we also get it in c9s/c10s? (We're hitting it there too: https://github.com/openshift/os/issues/1656).

PR #511, opened 2025-02-24
 conditionally remove tests from cross build

n80fr1n60:
in a cross build, test cannot be run, as target and host is different, 
so we remove it if we are doing a cross build

-> this is RFC - if it can be resolved another way, let me know

### DIFF
--- a/src/pins/tang/meson.build
+++ b/src/pins/tang/meson.build
@@ -1,11 +1,13 @@
 curl = find_program('curl', required: false)
 
-if curl.found()
+if curl.found() or meson.is_cross_build()
   bins += join_paths(meson.current_source_dir(), 'clevis-decrypt-tang')
   bins += join_paths(meson.current_source_dir(), 'clevis-encrypt-tang')
   mans += join_paths(meson.current_source_dir(), 'clevis-encrypt-tang.1')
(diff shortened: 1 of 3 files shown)


src/pins/tang/meson.build
```diff
@@ -1,11 +1,13 @@
 curl = find_program('curl', required: false)
 
-if curl.found()
+if curl.found() or meson.is_cross_build()
```
n80fr1n60, line 3: locally the cross build sysroot fs does not have "curl" command but embedded device has curl
-> recommendation for a cleaner solution ?

sergio-correia: I think this looks okay. Not great, but not too bad either. @sarroutbi: any opinions here?

sarroutbi: LGTM. Thanks

sarroutbi (approved):

n80fr1n60:
@sergio-correia @sergio-correia - if time permits, please have a look 
this one is related to my other one which was merged yesterday 
(https://github.com/latchset/tang/pull/153) )

sergio-correia (approved):
Thanks.

PR #513, opened 2025-03-13
Fix packages to install

sarroutbi:


sergio-correia (commented):
Looks good, but I would avoid the `-y`. Not a great practice to do these things without confirmation. While most people would do it sometimes, I would not recommend it in the README.

sergio-correia (approved):

sarroutbi:
Thanks for the quick review

PR #514, opened 2025-03-14
Remove dependencies and -y parameter

sarroutbi:
clevis-pin-pkcs11 has been changed to retrieve
its whole dependencies (socat, openssl, ...)
so it is not necessary to install them manually


sergio-correia (approved):

sarroutbi:
Thanks for the quick review

PR #522, opened 2025-06-06
fix missing pkcs11 when libexecdir is configured

ClaudiaJ:
should resolve error in dracut installing `clevis-luks-pkcs11-askpass` & `clevis-luks-pkcs11-askpin` for distros where `libexecdir` is configured elsewhere than `/usr/libexec` (Arch et. al.)


sarroutbi:
@ClaudiaJ : Sorry for letting this unmerged. I wanted to do some smoke testing before. Thanks for your PR !!!

PR #527, opened 2025-07-23
clevis-luks-askpass: allow non-block-device volumes

### LINKED ISSUE #526: Unable to unlock a file-based LUKS volume using sss (tang) pin at boot time
I have a LUKS volume inside a file `/root/.local/vault.img` that I can manually luksOpen / luksClose and mount without any issue.

I have the following packages installed (on Debian 12 Stable / amd64):

```
ii  clevis                          19-2                           amd64        automated encryption framework
ii  clevis-luks                     19-2                           amd64        LUKS integration for clevis
ii  clevis-systemd                  19-2                           amd64        systemd integration for clevis
```

I have added an sss pin to this volume using

```
clevis luks bind -d /root/.local/vault.img sss '{"t":1,"pins":{"tang":[{"url":"http://10.0.0.28:2828"}]}}'
```

and the tang server is reachable over network (confirmed by curl).

After this, I can successfully unlock the volume manually without a password using:

```
clevis luks unlock -d /root/.local/vault.img -n vault
```

I have added the following lines into /etc/crypttab

```
vault /root/.local/vault.img none luks
```

And into /etc/fstab

```
/dev/mapper/vault /root/vault ext4 noatime 0 2
```

after which, `systemctl daemon-reload` was issued to re-generate the mount unit files. 

However after all this, the system still keeps asking for manual entry of the passphrase during boot time, and it is unclear to me what exactly is wrong.

Any help would be much appreciated.

**EDIT:**

After much more digging I was able to get this setup into some sort of working state, by solving several problems:

**Problem 1.** mountpoints generated by systemd-cryptsetup-generator attempt to mount the volume before the network is online

**Solution.** Add _netdev option (documented in `man 8 mount`) to the /etc/fstab record and do `systemctl daemon-reload`. Working fstab record is:
```
/dev/mapper/vault /root/vault ext4 defaults,noatime,_netdev 0 2
```

**Problem 2.** `clevis-luks-askpass` binary explicitly checks if the volume that is being unlocked is a block special file (https://github.com/latchset/clevis/blob/master/src/luks/systemd/clevis-luks-askpass.in#L53). Obviously for my use case `/root/.local/vault.img` is just a normal file, so the check fails.

**Solution.** Comment out the `[ -b "${d}" ] || continue` line from `clevis-luks-askpass`. The fact that various checks in this code fail silently seriously affects the user experience and still remains a problem.

After this the system is able to boot and mount the volume automatically. I am not sure this is a good solution, perhaps there are ways to improve it.

x-qq:
Change the check on the volume that's being unlocked from -b (checks for block device) to -e (checks if file exists). This allows the code to work on LUKS volumes that are stored in normal files.

Resolves #526.

### DIFF
--- a/src/luks/systemd/clevis-luks-askpass.in
+++ b/src/luks/systemd/clevis-luks-askpass.in
@@ -50,7 +50,7 @@ while true; do
             esac
         done < "$question"
 
-        [ -b "${d}" ] || continue
+        [ -e "${d}" ] || continue
         [ -S "${s}" ] || continue
 
         if ! pt="$(clevis_luks_unlock_device "${d}")" || [ -z "${pt}" ]; then


sergio-correia (approved):
Thanks, LGTM

PR #533, opened 2025-10-05
Use bash instead of sh for clevis-luks-unlocker

Nicop06:
This is needed as clevis-luks-unlocker sources
clevis-luks-common-functions which requires bash due to use of `[[ ]]`.


oldium:
FYI - I have removed the need of having `clevis-luks-unlocker` in #462, so this might not be necessary when my PR gets merged.

oldium:
Outdated as #462 has been merged

Nicop06:
Thanks for letting me know.

PR #544, opened 2026-02-25
Correct LUKS hash handling

sarroutbi:
When a LUKS device is formatted with a non-default hash algorithm (e.g. sha512), clevis operations that add or update key slots (bind, regen, edit) did not pass the original hash to cryptsetup, causing new slots to silently default to sha256.
    
Add clevis_luks_get_hash() to extract the device's hash algorithm from the LUKS header (supports both LUKS1 "Hash spec:" and LUKS2 "Hash:" fields). Use it in clevis_luks_add_key() and clevis_luks_update_key() to pass --hash to cryptsetup, preserving the original hash algorithm.
    
Add unit tests for bind, regen and edit operations on both LUKS1 and LUKS2 devices formatted with sha512, verifying the hash is preserved through each operation.

### DIFF
--- a/src/luks/tests/regen-hash-luks1
+++ b/src/luks/tests/regen-hash-luks1
@@ -73,10 +73,10 @@ if [ "${enabled}" -ne 1 ]; then
     error "${TEST}: we should have only one slot enabled (${enabled})."
 fi
 
+SLT=1
 old_key=$(clevis_luks_unlock_device_by_slot "${DEV}" "${SLT}")
 
 # Now let's try regen.
-SLT=1
 if ! clevis luks regen -q -d "${DEV}" -s "${SLT}"; then
     error "${TEST}: clevis luks regen failed"
 fi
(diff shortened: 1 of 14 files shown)


src/luks/tests/regen-hash-luks1
```diff
@@ -0,0 +1,101 @@
...
+    error "${TEST}: error removing the default password from ${DEV}."
+fi
+
+# Making sure we have a single slot enabled.
+enabled=$(clevis_luks_used_slots "${DEV}" | wc -l)
+if [ "${enabled}" -ne 1 ]; then
+    error "${TEST}: we should have only one slot enabled (${enabled})."
+fi
+
+old_key=$(clevis_luks_unlock_device_by_slot "${DEV}" "${SLT}")
```
sergio-correia, line 77: It seems you are using SLT here before it has been initialized, in e.g. line 79 down below? Same thing with the luks2 test.

sarroutbi: Nice catch. Let me fix it

sergio-correia (approved):
LGTM, thanks

PR #549, opened 2026-04-16
dracut: fix running with v110 Dracut (#545)

### LINKED ISSUE #545: Ordering issue with dracut 110
Dracut 110 [changed the ordering of systemd-crypsetup to 71](https://github.com/dracut-ng/dracut-ng/pull/2111). This seems to make stuff not work anymore:

```
dracut[I]: *** Including module: clevis ***
Failed to add dependency on unit: Unit cryptsetup.target does not exist
```

Readding `cryptsetup.target` to `inst_multiple` (like in d7a2f38ed966a49f1204033efcb3c70d09c6641c) seems to fix the issue, but I'm not sure if this is the correct approach.

Tested on a recent daily build of Ubuntu Server 26.04.

```
ii  clevis-dracut                     21-1                                       amd64        Clevis integration for dracut
ii  dracut                            110-3ubuntu1                               all          Initramfs generator using udev
```

```
# ls -1Fv /usr/lib/dracut/modules.d/ | grep -E 'clevis|cryptsetup'
60clevis/
60clevis-pin-null/
60clevis-pin-sss/
60clevis-pin-tang/
60clevis-pin-tpm2/
71systemd-cryptsetup/
```

oldium:
Partinal revert of 0e3a8b60 ("dracut: fix running with pre-v103 Dracut")

The installation of `cryptsetup.target` by Dracut was originally made by `00systemd`, later `10systemd`, but was removed in [v104][1]. The installation somehow worked, because before v104 and after v104 there always was a module installing `cryptsetup.target` before `clevis` was installed - `systemd-cryptsetup`, which was moved from `90systemd-cryptsetup` to `01systemd-cryptsetup` in v105 and moved to `11systemd-cryptsetup` in v108.

According to that the version v104 was broken and now version v110 is broken again, because [v110][2] moved `systemd-cryptsetup` to `71systemd-cryptsetup`, which is installed after `50clevis`.

Fix this mess by returning back the installation of `cryptsetup.target` into `50clevis` module.

This fixes the following error during Dracut installation:

```
dracut[I]: *** Including module: clevis ***
Failed to add dependency on unit: Unit cryptsetup.target does not exist
```

[1]: https://github.com/dracut-ng/dracut-ng/commit/ad520855113d99d7551a7ab6c19870736f72cbc4
[2]: https://github.com/dracut-ng/dracut-ng/commit/d7bdbbb702abcd9c141decd0160284ad22a8b253

Fixes: #545

### DIFF
--- a/src/luks/dracut/clevis/module-setup.sh.in
+++ b/src/luks/dracut/clevis/module-setup.sh.in
@@ -38,6 +38,7 @@ install() {
         inst_multiple \
             $systemdsystemunitdir/clevis-luks-askpass.service \
             $systemdsystemunitdir/clevis-luks-askpass.path \
+            $systemdsystemunitdir/cryptsetup.target \
             @SYSTEMD_REPLY_PASS@ \
             @libexecdir@/clevis-luks-askpass
 


sergio-correia (approved):
Thanks for the fix!

PR #550, opened 2026-04-20
udisks2: fix LUKS2 auto-unlock failure with TPM2 pin

sergio-correia:
`token_to_jwe()` had an off-by-one error: `pkt->used--` was meant to exclude the null terminator from the byte count, but `snprintf()` already excludes it. This truncated the last byte of the JWE authentication tag before piping it to `clevis decrypt`.

The Rust `clevis-pin-tpm2` strictly validates base64url encoding and rejected the truncated tag ("Encoded text cannot have a 6-bit remainder"), causing TPM2 auto-unlock via udisks2 to always fail on LUKS2 devices. Tang was unaffected because the C jose library tolerates truncated base64url input.

### DIFF
--- a/src/luks/udisks2/test-token-to-jwe.c
+++ b/src/luks/udisks2/test-token-to-jwe.c
@@ -1,0 +1,133 @@
+/* vim: set tabstop=8 shiftwidth=4 softtabstop=4 expandtab smarttab colorcolumn=80: */
+/*
+ * Copyright (c) 2026 Red Hat, Inc.
+ * Author: Sergio Correia <scorreia@redhat.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


src/luks/udisks2/test-token-to-jwe.c
```diff
@@ -0,0 +1,133 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "token-to-jwe.h"
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
```
sarroutbi, line 25: Is this really being used? Can you give it a try to confirm it?

sarroutbi (commented):
Minor comment (https://github.com/latchset/clevis/pull/550/changes#diff-eeb31586df586f1ee075b14e094ab501873bafeb263cc8eb939d214748141030R25).

Rest of changes LGTM

src/luks/udisks2/test-token-to-jwe.c
sergio-correia: By "this" do you mean these 4 includes?

sarroutbi: No, sorry, just `string.h`

sergio-correia: `strlen()` and `strcmp()` come from `string.h`.

sarroutbi (approved):
LGTM

PR #552, opened 2026-04-28
Fix pkcs11 pubkey id parsing

sarroutbi:


src/pins/pkcs11/clevis-encrypt-pkcs11
```diff
@@ -99,7 +99,7 @@ fi
 PKEY="${CLEVIS_PKCS11}/pubkey"
 if ! id=$(pkcs11-tool ${slot_opt} -O 2>${ERR} ${module_opt} \
           | grep -i 'Public' -A10 2>${ERR} | grep 'ID:' \
-	  | head -1 | awk -F 'ID:' '{print $2}' | tr -d ' '); then
+	  | head -1 | awk -F 'ID:' '{print $2}' | awk '{print $1}'); then
```
sergio-correia: The new pkcs11-tool format (OpenSC 0.27.1, https://github.com/OpenSC/OpenSC/commit/c4abf2084) for small CKA_IDs (≤4 bytes) is `DECIMAL (0xHH:HH)`. `awk '{print $1}'` extracts the decimal part, but `pkcs11-tool --read-object --id` expects hex (`sc_hex_to_bin()` on the receiving end).

This works by accident for single-byte IDs 0x00–0x09 (where decimal == hex), but breaks for anything else. You can reproduce with SoftHSM on Fedora (OpenSC 0.27.1):

```
export SOFTHSM2_CONF=/tmp/softhsm2.conf
mkdir -p /tmp/tokens
echo "directories.tokendir = /tmp/tokens" > $SOFTHSM2_CONF
P11LIB=/usr/lib64/pkcs11/libsofthsm2.so

softhsm2-util --init-token --free --label test --so-pin 12345678 --pin 123456
pkcs11-tool --module $P11LIB --login --pin 123456 \
    --keypairgen --key-type RSA:2048 --id 0a --label testkey

# Shows "ID: 10 (0x0a)" — decimal, not hex
pkcs11-tool --module $P11LIB -O

# The fix extracts "10" (decimal)
pkcs11-tool --module $P11LIB -O \
    | grep -i 'Public' -A10 | grep 'ID:' \
    | head -1 | awk -F 'ID:' '{print $2}' | awk '{print $1}'

# Fails: looks for CKA_ID {0x10}, not {0x0a}
pkcs11-tool --module $P11LIB --read-object --type pubkey --id 10

# Works: correct CKA_ID
pkcs11-tool --module $P11LIB --read-object --type pubkey --id 0a
```

The fix should extract the hex from inside the parentheses when present, falling back to the raw value for old format. Something like:

```bash
| head -1 | awk -F 'ID:' '{
    val = $2
    if (match(val, /\(0x([^)]+)\)/, m)) {
        print m[1]
    } else {
        gsub(/^[[:space:]]+|[[:space:]]+$/, "", val)
        print val
    }
}'
```

src/pins/pkcs11/tests/pkcs11-id-parsing
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash -xe
+#
+# Copyright (c) 2024 Red Hat, Inc.
```
sergio-correia: 2026?

sergio-correia (changes requested):

sergio-correia (approved):
LGTM, thanks!

PR #557, opened 2026-05-12
Add meson build option to make tpm1 pin optional

sarroutbi:
Add a 'tpm1' feature option (auto/enabled/disabled) to allow distros to disable TPM 1.2 support at build time. When disabled, none of the tpm1 pin sources, tests, dracut modules, systemd units, or LUKS helper functions are built or installed.

The default is 'auto' which preserves existing behavior: build tpm1 if dependencies (tpm_sealdata, tpm_unsealdata) are found, skip gracefully otherwise.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -63,3 +63,5 @@ if a2x.found()
 else
   warning('Will not build man pages due to missing dependencies!')
 endif
+
+summary('TPM 1.2 pin', get_option('tpm1'), section: 'Features')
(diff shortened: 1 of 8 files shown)


prikryla:
/packit test --identifier pkcs11 --labels centos-stream-9-x86_64

sarroutbi:
/packit test --identifier pkcs11 --labels centos-stream-9-x86_64

sergio-correia (commented):
Thanks, this looks good overall, but a couple of nits:

- We have guards in clevis-tcsd.conf installation but not in `clevis-luks-askpass.service.in` itself, which unconditionally lists
```
  After=tcsd.service
  Wants=tcsd.service
```
- It would be nice to add a `summary()` call so packagers can confirm the tpm1 option state at configure time, e.g. `summary('TPM 1.2 pin', get_option('tpm1'), section: 'Features')`

sergio-correia (approved):

PR #561, opened 2026-05-31
dracut: fix some portability issues in non-systemd environments

oxiez:
Fixes some issues I ran into when trying to use Clevis in Void Linux:
- Missing `chmod` install in my init environment but is used by unlocker scripts.
- `systemd-reply-password` path was required during build time. It falls back to empty string since it's not used with the dracut hooks path.

Thanks.

### DIFF
--- a/src/luks/dracut/clevis/meson.build
+++ b/src/luks/dracut/clevis/meson.build
@@ -5,7 +5,7 @@ if dracut.found()
 
   dracut_data = configuration_data()
   dracut_data.merge_from(data)
-  dracut_data.set('SYSTEMD_REPLY_PASS', sd_reply_pass.path())
+  dracut_data.set('SYSTEMD_REPLY_PASS', sd_reply_pass.found() ? sd_reply_pass.path() : '')
 
   configure_file(
     input: 'module-setup.sh.in',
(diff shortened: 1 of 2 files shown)


louiszn:
I independently tested this PR while updating the Void Linux `clevis` package from 20 to 23.

Test environment:

```
Void Linux x86_64
runit / non-systemd
dracut 109
clevis 23
TPM2-bound LUKS root
```

Before applying these patches, Clevis 23 failed to configure on this non-systemd setup because `systemd-reply-password` was not available:

```
Run-time dependency systemd found: NO
Program ... systemd-reply-password found: NO
src/luks/systemd/meson.build:55: WARNING: Will not install systemd support due to missing dependencies!
Run-time dependency dracut found: YES 109
src/luks/dracut/clevis/meson.build:8: ERROR: Unable to get the path of a not-found external program
```

I also reproduced the missing `chmod` issue during early boot after TPM2 auto-unlock:

```
/bin/clevis-password-unlocker-prepare: line 25: chmod: command not found
/bin/clevis-password-unlocker: line 183: chmod: command not found
```

After applying this PR's patches, Clevis 23 configures and builds successfully on Void Linux, the dracut module is generated correctly without `systemd-reply-password`, and TPM2-bound LUKS auto-unlock continues to work. `chmod` is now included in the initramfs and the early-boot warning disappears.

So this PR fixes the issue for my Void Linux non-systemd/dracut test case as well.

sergio-correia (approved):
Thanks!

PR #562, opened 2026-06-18
dracut/tang: Order clevis-luks-askpass after network-online.target and more fixes

rmetrich:
When Tang bindings are present, `rd.neednet=1` is set to bring up the network in the initramfs, but `clevis-luks-askpass.path` has no ordering dependency on `network-online.target`. This means systemd can activate the askpass path unit before the network is ready, causing curl to fail when contacting the Tang server:

~~~
Jun 18 10:32:39 vm-clevis10 clevis-luks-askpass[760]: Error communicating with server http://vm-tang10
~~~

Install a systemd drop-in for `clevis-luks-askpass.path` that adds `After=network-online.target` and `Wants=network-online.target`, ensuring the network is online before clevis attempts to decrypt via Tang.

### DIFF
--- a/src/luks/dracut/clevis-pin-tang/module-setup.sh.in
+++ b/src/luks/dracut/clevis-pin-tang/module-setup.sh.in
@@ -25,9 +25,10 @@ depends() {
 
 have_tang_bindings() {
     . clevis-luks-common-functions
-    local dev
+    local dev bindings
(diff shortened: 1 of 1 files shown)


sergio-correia (approved):
Thank you, it looks good to me.

PR #566, opened 2026-07-07
Add optional mTLS support to the tang pin

liyue32:
### Summary

Adds optional mutual-TLS (mTLS) support to the Tang pin, allowing Clevis to authenticate to a Tang server (typically fronted by a TLS-terminating reverse proxy) with a client certificate, and to verify the server certificate against a custom CA.
The feature is fully opt-in: when none of the new properties are set, the pin behaves exactly as before (plain HTTP/HTTPS with the system's default CA trust). Existing JWEs and workflows are unaffected.

### Motivation

Adding mTLS lets the Tang server (typically via a TLS-terminating reverse proxy) authenticate the client's identity through its client certificate. This unlocks per-machine access control on top of the existing network-binding model. In particular, single-machine revocation: revoking or refusing a specific client certificate immediately prevents that one machine from recovering its keys, without affecting any other machine or requiring key rotation on the server. This matches how we want to adopt Clevis out of the box at Meta: bind machines to Tang for automated unlock, and retain the ability to revoke an individual host's recovery access by its identity. It's also useful more generally for any deployment where the Tang server is fronted by HTTPS with client-certificate authentication, or uses a server certificate signed by a private/internal CA.

### What changed
Three new optional config properties are accepted by clevis encrypt tang:
1. cacert: curl --cacert
2. cert: curl --cert
3. key: curl --key 

When any of above are set: 
1. passed to curl when fetching the advertisement during encryption, and
2. persisted into the JWE protected header under clevis.tang, so that clevis decrypt reuses them for the POST /rec request without any additional configuration (Clevis decryption takes no config by design, so this metadata must travel in the JWE).

Example:
```
  clevis encrypt tang '{
    "url":"https://tang.example.com",
    "cacert":"/etc/clevis/ca.crt",
    "cert":"/etc/clevis/client.crt",
    "key":"/etc/clevis/client.key"
  }' < PT > JWE
```

### Backward compatibility
- When no mTLS properties are set, the injected curl option array is empty, so curl is invoked identically to before.
- Existing JWEs have no mTLS header fields; decryption reads them as absent and takes the unchanged code path.

### Tests
- Adds a new pin-tang-mtls integration test plus tang_generate_certs, tang_run_mtls, and tang_mtls_sanity_check helpers. The harness fronts tangd with a TLS listener (socat OPENSSL-LISTEN) that mandates client-certificate verification. The test covers:
    - Positive: encrypt (GET /adv) + decrypt (POST /rec) round-trip succeed over mTLS. Since the server enforces client-cert verification, success proves the pin actually presents the client certificate.
  - Header round-trip: the cacert/cert/key paths are correctly persisted into and read back from the JWE protected header.
  - Negative: without a client certificate, encryption fails — proving mTLS is enforced, not silently downgraded.

### DIFF
--- a/src/pins/tang/clevis-decrypt-tang
+++ b/src/pins/tang/clevis-decrypt-tang
@@ -87,6 +87,25 @@ if ! url="$(jose fmt -j- -Og clevis -g tang -g url -Su- <<< "$jhd")"; then
     exit 1
 fi
 
+# Optional (m)TLS parameters. Absent for JWEs created without mTLS, in which
+# case curl behaves exactly as before.
+cacert="$(jose fmt -j- -Og clevis -g tang -g cacert -Su- <<< "$jhd")" || true
+cert="$(jose fmt -j- -Og clevis -g tang -g cert -Su- <<< "$jhd")" || true
+key="$(jose fmt -j- -Og clevis -g tang -g key -Su- <<< "$jhd")" || true
+
+# Fail with a targeted diagnostic if a referenced (m)TLS file is missing.
+# Otherwise curl -sfg fails silently below and the generic "Error
+# communicating with server" sends the operator chasing network issues,
+# when the real cause is a rotated cert or a file left out of the initramfs.
+[ -n "$cacert" ] && [ ! -f "$cacert" ] && { echo "mTLS CA cert '$cacert' not found (is it in the initramfs?)" >&2; exit 1; }
+[ -n "$cert" ]   && [ ! -f "$cert" ]   && { echo "mTLS client cert '$cert' not found (is it in the initramfs?)" >&2; exit 1; }
+[ -n "$key" ]    && [ ! -f "$key" ]    && { echo "mTLS client key '$key' not found (is it in the initramfs?)" >&2; exit 1; }
+
+curl_opts=()
+[ -n "$cacert" ] && curl_opts+=(--cacert "$cacert")
+[ -n "$cert" ]   && curl_opts+=(--cert "$cert")
+[ -n "$key" ]    && curl_opts+=(--key "$key")
+
 if ! crv="$(jose fmt -j- -Og crv -Su- <<< "$clt")"; then
     echo "Unable to determine EPK's curve!" >&2
     exit 1
@@ -101,7 +120,7 @@ xfr="$(jose jwk exc -i '{"alg":"ECMR"}' -l- -r- <<< "$clt$eph")"
 
 rec_url="$url/rec/$kid"
 ct="Content-Type: application/jwk+json"
-if ! rep="$(curl -sfg -X POST -H "$ct" --data-binary @- "$rec_url" <<< "$xfr")"; then
+if ! rep="$(curl -sfg "${curl_opts[@]}" -X POST -H "$ct" --data-binary @- "$rec_url" <<< "$xfr")"; then
     echo "Error communicating with server $url" >&2
     exit 1
 fi
--- a/src/pins/tang/clevis-encrypt-tang
+++ b/src/pins/tang/clevis-encrypt-tang
@@ -45,6 +45,16 @@ if [ -t 0 ]; then
     echo "  adv: <string>   A filename containing a trusted advertisement"
     echo "  adv: <object>   A trusted advertisement (raw JSON)"
     echo
+    echo "  cacert: <string>  Path to a CA bundle used to verify the Tang"
+    echo "                    server's TLS certificate (curl --cacert)"
+    echo
+    echo "  cert: <string>    Path to the client TLS certificate for mutual"
+    echo "                    TLS authentication (curl --cert)"
+    echo
+    echo "  key: <string>     Path to the private key for the client TLS"
+    echo "                    certificate (curl --key). May be omitted if the"
+    echo "                    key is bundled in the 'cert' file"
+    echo
     echo "Obtaining the thumbprint of a trusted signing key is easy. If you"
     echo "have access to the Tang server's database directory, simply do:"
     echo
@@ -77,6 +87,37 @@ fi
 
 thp="$(jose fmt -j- -Og thp -Su- <<< "$cfg")" || true
 
+# Optional (m)TLS parameters. If none are set, curl behaves exactly as before
+# (plain HTTP/TLS, no client certificate and default CA verification).
+cacert="$(jose fmt -j- -Og cacert -Su- <<< "$cfg")" || true
+cert="$(jose fmt -j- -Og cert -Su- <<< "$cfg")" || true
+key="$(jose fmt -j- -Og key -Su- <<< "$cfg")" || true
+
+# Validate the (m)TLS file paths up front, matching the 'adv' check below.
+# Otherwise a typo'd path either surfaces as a misleading "Unable to fetch
+# advertisement" error, or (when 'adv' is inline and curl is never called)
+# gets baked into the JWE header and only fails later at decrypt time.
+[ -n "$cacert" ] && [ ! -f "$cacert" ] && { echo "CA cert file '$cacert' not found!" >&2; exit 1; }
+[ -n "$cert" ]   && [ ! -f "$cert" ]   && { echo "Client cert file '$cert' not found!" >&2; exit 1; }
+[ -n "$key" ]    && [ ! -f "$key" ]    && { echo "Client key file '$key' not found!" >&2; exit 1; }
+
+# 'key' on its own is a no-op: curl ignores --key unless --cert is also given.
+[ -n "$key" ] && [ -z "$cert" ] && { echo "'key' requires 'cert' to also be set!" >&2; exit 1; }
+
+# Reject http:// when cacert or cert is set: curl would silently ignore them.
+if [ -n "$cacert" ] || [ -n "$cert" ]; then
+    case "$url" in
+        https://*) ;;
+        *) echo "TLS options (cacert/cert/key) require an https:// URL!" >&2
+           exit 1 ;;
+    esac
+fi
+
+curl_opts=()
+[ -n "$cacert" ] && curl_opts+=(--cacert "$cacert")
+[ -n "$cert" ]   && curl_opts+=(--cert "$cert")
+[ -n "$key" ]    && curl_opts+=(--key "$key")
+
 ### Get the advertisement
 if jws="$(jose fmt -j- -g adv -Oo- <<< "$cfg")"; then
     thp="${thp:-any}"
@@ -92,7 +133,7 @@ elif jws="$(jose fmt -j- -g adv -Su- <<< "$cfg")"; then
     fi
 
     thp="${thp:-any}"
-elif ! jws="$(curl -sfg "$url/adv/$thp")"; then
+elif ! jws="$(curl -sfg "${curl_opts[@]}" "$url/adv/$thp")"; then
     echo "Unable to fetch advertisement: '$url/adv/$thp'!" >&2
     exit 1
 fi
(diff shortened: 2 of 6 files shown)


liyue32:
@sarroutbi I could not see the detailed error message of centos-stream-10-x86_64:pkcs11 failure, but seems like it is timeout issue, could you help to take a look if this is related with my PR changes?

src/pins/tang/tests/pin-tang-mtls
```diff
@@ -0,0 +1,69 @@
...
+#
+
+. tang-common-test-functions
+
+on_exit() {
+    exit_status=$?
+    tang_stop "${TMP}"
+    [ -d "$TMP" ] && rm -rf "$TMP"
+    exit "${exit_status}"
+}
```
sergio-correia, line 27: The `on_exit` trap is registered (line 29) before `tang_mtls_sanity_check` (line 32), which runs before `TMP` is assigned (line 34). If the sanity check calls `skip_test`, the trap fires with `TMP` unset, and `tang_stop ""` ends up calling `error "tang_stop: please specify 'basedir'"` which exits 1. That converts a SKIP (exit 77) into a FAIL (exit 1).

This will break CI on systems that have socat+tangd but socat was built without OpenSSL support.

Compare with `pin-tang`, where `TMP="$(mktemp -d)"` is set right after the trap with no skip-able call in between.

Please guard the cleanup so it tolerates an unset `TMP`:

```suggestion
on_exit() {
    exit_status=$?
    [ -n "${TMP}" ] && tang_stop "${TMP}"
    [ -d "${TMP}" ] && rm -rf "$TMP"
    exit "${exit_status}"
}
```

src/pins/tang/clevis-encrypt-tang
```diff
@@ -77,6 +87,17 @@ fi
 
 thp="$(jose fmt -j- -Og thp -Su- <<< "$cfg")" || true
 
+# Optional (m)TLS parameters. If none are set, curl behaves exactly as before
+# (plain HTTP/TLS, no client certificate and default CA verification).
+cacert="$(jose fmt -j- -Og cacert -Su- <<< "$cfg")" || true
+cert="$(jose fmt -j- -Og cert -Su- <<< "$cfg")" || true
+key="$(jose fmt -j- -Og key -Su- <<< "$cfg")" || true
```
sergio-correia, line 94: These paths are never validated for existence before use. The existing `adv` file path *is* checked (line 105: `if ! [ -f "$jws" ]`), so there is a consistency gap.

If a user typos a path (e.g., `clietn.crt`), two things can happen:

1. When the advertisement is fetched via curl, `curl -sfg` fails silently and the user sees `"Unable to fetch advertisement"`, which points them at network/DNS issues instead of the real problem.

2. When `adv` is provided inline, curl is never called, encryption *succeeds*, and the bad path gets permanently baked into the JWE header. Every decrypt attempt then fails at boot with `"Error communicating with server"`.

Please add file-existence checks after extracting the paths, matching the `adv` validation pattern:

```bash
[ -n "$cacert" ] && [ ! -f "$cacert" ] && { echo "CA cert file '$cacert' not found!" >&2; exit 1; }
[ -n "$cert" ]   && [ ! -f "$cert" ]   && { echo "Client cert file '$cert' not found!" >&2; exit 1; }
[ -n "$key" ]    && [ ! -f "$key" ]    && { echo "Client key file '$key' not found!" >&2; exit 1; }
```

src/pins/tang/clevis-decrypt-tang
```diff
@@ -87,6 +87,17 @@ if ! url="$(jose fmt -j- -Og clevis -g tang -g url -Su- <<< "$jhd")"; then
     exit 1
 fi
 
+# Optional (m)TLS parameters. Absent for JWEs created without mTLS, in which
+# case curl behaves exactly as before.
+cacert="$(jose fmt -j- -Og clevis -g tang -g cacert -Su- <<< "$jhd")" || true
```
sergio-correia, line 92: During decryption, `cacert`, `cert`, `key` (and `url`) are all extracted from the JWE protected header and fed to curl (line 115) before the AEAD integrity check at line 128 (`jose jwe dec`). The `url` field was already consumed unauthenticated before this patch, so this is a pre-existing design property of clevis-tang. However, `cacert` qualitatively extends the attack surface: an attacker who can modify the stored JWE can now also override which CA curl trusts, enabling MITM against the legitimate tang server with a self-signed cert planted on disk. Previously, modifying `url` alone still required a certificate trusted by the system CA store.

The practical exploitability is limited (requires disk write access + network position, and the ECMR exchange value alone does not reveal the decryption key), so this may be acceptable as a known trade-off.

Worth considering: validate that cert/key/cacert paths are under an expected prefix (e.g., `/etc/clevis/`), or use indirection (store a profile name, resolve to paths at runtime from a fixed directory). At minimum, document the trust model explicitly so users understand the implication.

src/pins/tang/clevis-decrypt-tang
```diff
@@ -101,7 +112,7 @@ xfr="$(jose jwk exc -i '{"alg":"ECMR"}' -l- -r- <<< "$clt$eph")"
 
 rec_url="$url/rec/$kid"
 ct="Content-Type: application/jwk+json"
-if ! rep="$(curl -sfg -X POST -H "$ct" --data-binary @- "$rec_url" <<< "$xfr")"; then
+if ! rep="$(curl -sfg "${curl_opts[@]}" -X POST -H "$ct" --data-binary @- "$rec_url" <<< "$xfr")"; then
     echo "Error communicating with server $url" >&2
```
sergio-correia, line 124: When mTLS parameters are present in the JWE header but the referenced files no longer exist (cert rotation, initramfs rebuilt without them), `curl -sfg` fails silently and all the user sees is this generic message. In the initramfs at boot, this sends the operator chasing network/firewall issues when the real problem is a missing certificate file.

The man page already warns about ensuring files are in the initramfs, which tells us this is a known footgun. Would it make sense to check the files exist before the curl call and produce a targeted diagnostic?

```bash
[ -n "$cacert" ] && [ ! -f "$cacert" ] && { echo "mTLS CA cert '$cacert' not found (is it in the initramfs?)" >&2; exit 1; }
[ -n "$cert" ]   && [ ! -f "$cert" ]   && { echo "mTLS client cert '$cert' not found (is it in the initramfs?)" >&2; exit 1; }
[ -n "$key" ]    && [ ! -f "$key" ]    && { echo "mTLS client key '$key' not found (is it in the initramfs?)" >&2; exit 1; }
```

src/pins/tang/clevis-encrypt-tang
```diff
@@ -77,6 +87,17 @@ fi
...
+
+curl_opts=()
+[ -n "$cacert" ] && curl_opts+=(--cacert "$cacert")
+[ -n "$cert" ]   && curl_opts+=(--cert "$cert")
+[ -n "$key" ]    && curl_opts+=(--key "$key")
```
sergio-correia, line 119: Specifying `key` without `cert` is a no-op: curl ignores `--key` when `--cert` is not also provided. If the server requires mTLS, the handshake fails with a generic error and there is no indication that `cert` was missing. Please add a consistency check:

```bash
[ -n "$key" ] && [ -z "$cert" ] && { echo "'key' requires 'cert' to also be set!" >&2; exit 1; }
```

src/pins/tang/clevis-decrypt-tang
liyue32: Thanks a lot for your review! In my understanding, if attacker could modify JWE header, they could probably also modify the ca trust store in /etc/pki. If this is correct, cacert does not add more security risk here (correct me if my understanding is wrong). I also thought about adding validation for cert/key/cacert paths, but considering this aims to be used generically, it is hard to add any specific cacert path validation so I feel path-prefix validation felt too restrictive here. 
I added one section named "mTLS SECURITY MODEL" in clevis-encrypt-tang.1.adoc to describe the potential risk of `cacert` and recommende that the CA be installed in the system trust store (with cacert omitted) where possible

liyue32 (commented):
address the comments by 
* add cert, key and cacert path validation 
* Add security model document of using cacert

liyue32:
@sarroutbi @sergio-correia I addressed all the previous comments, would you mind taking a look again? Thanks!

src/pins/tang/clevis-encrypt-tang.1.adoc
```diff
@@ -76,6 +76,50 @@ This command uses the following configuration properties:
...
+== mTLS EXAMPLE
+
+    $ cfg='{"url":"https://tang.srv","cacert":"/etc/clevis/ca.crt",
+            "cert":"/etc/clevis/client.crt","key":"/etc/clevis/client.key"}'
+    $ clevis encrypt tang "$cfg" < PT > JWE
+
+== mTLS SECURITY MODEL
+
+The *cacert*, *cert* and *key* properties are stored as plaintext in the JWE
+header without authentication, so there is a chance they could be modified.
```
sergio-correia: The wording here says the mTLS properties are stored "without authentication", but they're placed in the JWE `protected` header (lines 193-195 of the encrypt script), which is the AAD for AES-256-GCM per RFC 7516 Section 5.1. Modifying any field in the protected header causes `jose jwe dec` to reject the token.

The real subtlety is that `clevis-decrypt-tang` reads and acts on these values (the curl request at line 123) *before* the AEAD integrity check runs (line 136). So tampering is detected, but only after one network request has already been made with the tampered parameters.

Would something like this be more accurate?

```suggestion
The *cacert*, *cert* and *key* properties are stored in the JWE *protected*
header, which is integrity-protected by the AES-256-GCM authentication tag.
Modifying any of these values will cause decryption to fail. However, the
integrity check is only performed *after* the values have been used to contact
the Tang server (the decryption key is needed to verify the tag, and the Tang
server must be contacted to derive that key). This means a tampered JWE will
cause the system to fail to unlock, but one curl request will have been sent
using the tampered parameters.
```

src/pins/tang/clevis-encrypt-tang
```diff
@@ -77,6 +87,28 @@ fi
...
+# Validate the (m)TLS file paths up front, matching the 'adv' check below.
+# Otherwise a typo'd path either surfaces as a misleading "Unable to fetch
+# advertisement" error, or (when 'adv' is inline and curl is never called)
+# gets baked into the JWE header and only fails later at decrypt time.
+[ -n "$cacert" ] && [ ! -f "$cacert" ] && { echo "CA cert file '$cacert' not found!" >&2; exit 1; }
+[ -n "$cert" ]   && [ ! -f "$cert" ]   && { echo "Client cert file '$cert' not found!" >&2; exit 1; }
+[ -n "$key" ]    && [ ! -f "$key" ]    && { echo "Client key file '$key' not found!" >&2; exit 1; }
+
+# 'key' on its own is a no-op: curl ignores --key unless --cert is also given.
+[ -n "$key" ] && [ -z "$cert" ] && { echo "'key' requires 'cert' to also be set!" >&2; exit 1; }
```
sergio-correia, line 105: When `cacert`, `cert`, or `key` are set but `url` uses `http://`, curl silently ignores `--cacert`, `--cert`, and `--key`. Encryption succeeds, the paths get baked into the JWE header, but all traffic goes over plaintext. The user thinks mTLS is protecting the connection when it isn't.

Would it make sense to reject this at bind time, similar to the `key`-without-`cert` check right above?

```bash
if [ -n "$cacert" ] || [ -n "$cert" ]; then
    case "$url" in
        https://*) ;;
        *) echo "TLS options (cacert/cert/key) require an https:// URL!" >&2
           exit 1 ;;
    esac
fi
```

sergio-correia (approved):
LGTM, thanks!

PR #569, opened 2026-07-14
build: Non-automagic pkcs11 pin support

salahcoronya:
All the pins (except tpm1) use automagic detection. Most of the other pins this isn't really a problem, as their dependencies are small and/or already likely to be installed. However, the pkcs11 pin pulls in pcsc-lite (which isn't a very common package), which is turns pulls in polkit (which some people consider bloat) and all its dependencies. 

The others are either have low dependency requirements, or are highly desirable (tpm2), and can simply be made outright requirements by the packager.


sarroutbi:
@LGTM. @sergio-correia , WDYT? We should take into account enabling the feature in next Fedora releases in case this is merged...

sarroutbi (approved):
LGTM

sergio-correia (approved):
LGTM, thanks

sergio-correia:
> @lgtm. @sergio-correia , WDYT? We should take into account enabling the feature in next Fedora releases in case this is merged...

Yeah, we can enable it there in the next update we do.

